PR #2, opened 2020-02-13
Add complimentary HTML parser and test for commonmark compliance

chrisjsewell:


choldgraf:
You mean "renderer", yeah? And this is basically just extending the mistletoe renderer to include the extra syntax that we're supporting on the markdown side?

chrisjsewell:
Yeh the main purpose is just so that I can run the same text -> AST parser that I’m using for the docutils renderer, and make sure it will still parse standard Markdown, by using the HTML render to generate and test the ~500 examples supplied by the commonmark-spec repo

PR #9, opened 2020-02-13
Add Target span token

chrisjsewell:
You can now do e.g.

```
(header_target)= 
# Header

{ref}`header_target`
```


choldgraf:
Nice! I'll add some docs...


Where did that syntax come from? It's a new one for me, is there prior art in pandoc or something?

PR #22, opened 2020-02-13
adding inter-doc links and more explanation of tokens

choldgraf:


chrisjsewell:
@choldgraf you missed that LineComments are an extension

chrisjsewell:
Also AutoLinks are not an extension

choldgraf:
Nice - good catch. Fixed both in the latest push

Something I just realized is that this PR will remove some information from the docs, which is the order of precedence for the extended syntax and the commonmark syntax. Can you think of a good place to insert that information back in here? I do think it's more valuable to explicitly highlight which syntax is new vs. standard...

choldgraf:
(depending on how you wanna resolve my last question, this PR is ready to go from me, but if you'd like me to add back in "precedence" information somehow (maybe a link in the code for where precedence is defined?) I'm happy to do so

chrisjsewell:
I’ll do this in a bit, just adding a FrontMatter token 😄

choldgraf:
@chrisjsewell I just rebased on `develop` and added in an internal ref for the frontmatter section 👍

chrisjsewell:
Just added a quick commit here, because I'm too lazy to do my own PR

choldgraf:
heh sounds good - ready to merge if tests are happy? If so then I will **squash and merge** because I'm not an animal.

choldgraf:
@chrisjsewell btw - how do you feel about just only allowing squash+merge for a bit?

chrisjsewell:
Yeh sounds good to me 👍

PR #25, opened 2020-02-13
Add support for FrontMatter block tokens (initial YAML block)

chrisjsewell:


choldgraf:
oh man I love this - I wonder if we can get clever and find a way to define `toctree:` information in the YAML block, rather than embedded in the content somewhere...

PR #28, opened 2020-02-13
added png fig

jstac:
Added missing png figure to docs folder


choldgraf:
Hmmm - it looks like the error is because sphinx isn't parsing "scale" properly...I see in the docs there's a space between the number and the % sign. @jstac wanna try adding that in and seeing if the docs build in that case?

so basically, we have:

```
scale: 80%
```
and we want

```
scale: 80 %
```

jstac:
Added space.  That's a bit fragile, isn't it?

choldgraf:
that works, thanks!

PR #36, opened 2020-02-14
Prototyping an IPYNB parser

choldgraf:
This is an incomplete attempt at an `.ipynb` parser in order to also try understanding how these various docutils elements work.

It doesn't currently work, but it has what I *think* is the pseudocode that would do what we want.

A few questions that I have for @chrisjsewell after looking through a bit:

* Is this structure the general structure you had in mind as well?
* The big "config spec" at the beginning of your class - is that something we need to copy into the IPynbParser class or does it inherit it?
* the `parse(` method doesn't actually return anything, so I assume that there's something in the context manager that tells Sphinx what it should take from the parser, is that right?
* It's a bit unclear what is a "Mistletoe" object and what is a "docutils" object...but is the way I'm structuring things here correct? I was imagining the `<document` parent would have a list of children, each corresponding to a `<container type="cell"` object.


chrisjsewell:
> Is this structure the general structure you had in mind as well?

Kind of, but you've gone wrong in a few places (see below).

> The big "config spec" at the beginning of your class - is that something we need to copy into the IPynbParser class or does it inherit it?

Yep if your inheriting from `MystParser` then that should be fine. that config is copied straight from the `RSRParser`, I don't if all of the options are actually needed, but will check at some point.

> the `parse(` method doesn't actually return anything, so I assume that there's something in the context manager that tells Sphinx what it should take from the parser, is that right?

The is adding to the input `document` (i.e. the docutils doctree AST) in-place, so nothing needs to be returned.

> It's a bit unclear what is a "Mistletoe" object and what is a "docutils"...

As above, the `document` is the root node of the docutils AST, and the renderer adds things to it and its (nested) descendants.
The mistletoe AST works a bit differently, in that it 'generates itself' when you initialise it.
So in `doc = Document("notebook")` you are actually saying 'parse the "notebook" string to AST' (i.e. its wrong 😆).

You need to build a separate mistletoe AST for each text cell string, then use that to add a child to the `document`:

```python
nb_cell = container(kind="cell", cell_type="markdown")
document += nb_cell
renderer = DocutilsRenderer(document=document, current_node=nb_cell)
with renderer:
    # the context manager temporarily sets some global parsing variables
    myst_ast = Document(cell['source'])
    renderer.renderer(myst_ast)
``` 

An extra thing I think would also be good, would be to override the `document.reporter.system_message` method (which is where parsing errors/warnings/info are reported), to report the cell number (as well as the standard line number).
See `docutils.utils.Reporter.system_message` for its implementation.

chrisjsewell:
FYI you should now use the SphinxRenderer` not the `DocutilsRenderer`

choldgraf:
OK cool, thanks for the help. I think the latest push fixes some of the parsing issues you had mentioned. It now correctly parses both the markdown cells and the *inputs* of the code cells. A few extra points:

* I'm still not sure what to do about the output cells...is there a quick stop-gap we can use so that we can iterate on this functionality?
* Maybe we want to use something like the Sphinx code-block directive instead of a `literal_block` node? That said, I am not sure how to get access to the list of available directives...
* If we add extra `kwarg` pairs to our `container` objects, will those propagate to the build phase? (`classes=` seems to be the way that extra classes are added at HTML render, but I guess we should still keep that information in another metadata key for other renderers to use)

BTW - you should also feel free to push to this branch. I don't want to block this feature getting in if I am being excruciatingly slow for you :-)

choldgraf:
OK this now has a hacky stop-gap solution for dealing with outputs. It totally re-uses the machinery from Jupyter Sphinx just to have something that works, but we should treat this as a pure stopgap to prototype functionality, it shouldn't be treated as a final fix at all.

chrisjsewell:
Mow implemented in `MyST-NB`

PR #37, opened 2020-02-14
viz helper

choldgraf:
This is a little visualization helper class to look at a docutils AST. It is a class that acts like a function, because that seemed like the best way to make it degrade gracefully if it was invoked outside of a Jupyter environment.

It basically takes an AST, converts it into a dictionary, and then uses the IPython `JSON` display function on it. If it's in a text environment, it just prettyprints the object.

No tests yet but I can write one if this is a reasonable addition...it's a little bit hacky so I'm not sure if it's useful or not...


choldgraf:
(correct me if you disagree, but after having used docutils a bit more now, I don't think this would actually be that useful...I'm gonna close this)

PR #45, opened 2020-02-15
reorganizing docs and using pydata theme

choldgraf:
I remember last week @chrisjsewell said he wasn't super pleased with the default "alabaster" theme. So this PR does two things:

* Updates the theme to the newer Pydata sphinx theme, which is a bit easier to navigate
* Rearranges the docs a little bit so that they are cleaner, and so they're easier to extend by adding content to sub-folders

Here's how it looks:

https://72-240151150-gh.circle-artifacts.com/0/html/index.html

What do folks think?

*note: I also added a dummy logo just to have something there*


choldgraf:
here's how it looks:

https://72-240151150-gh.circle-artifacts.com/0/html/index.html

jstac:
Thanks @choldgraf , it looks beautiful IMO.  

And `:doc:` is working too `:-)`

chrisjsewell:
It looks nice ta. The only thing missing though is a link from the examples, back to their source text

choldgraf:
OK cool - lemme merge this and then I'll add in links back to the source text in another PR...I'm on the bus now and documentation improvements always help me pass the time :-)

Issue #42, opened 2020-02-15
Explore using sphinx-jupyter 


choldgraf:
I was looking into sphinx-jupyter and I think it might be useful as a part of our build chain. It specifies directives for parsing and running cells as a doctree transformation, and I think expects people will directly use jupyter directives as opposed to using jupyter notebooks as inputs.

Three ways I could imagine using it (off top of my head) 

* if we end up doing execution within sphinx we could try utilizing this directly. In this case we might want to add some caching somehow 
* if we keep execution pre-sphinx we could try seeing if jupyter-sphinx would be friendly to allowing a read-only ipynb parser that uses the myst markdown parser under the hood.
* if they aren't OK with that, we can take heavy inspiration or reuse components from jupyter sphinx in how we render cells, outputs, etc. 

I wonder if @akhmerov has thoughts on that!

chrisjsewell:
Not to pick a fight lol, but I feel we are kind of superseding this package. So the `take heavy inspiration` option seems the way forward for me.

chrisjsewell:
FYI @choldgraf if you could label new issues and (if applicable) add them to the milestone 😁

choldgraf:
I think it somewhat depends on what functionality is inside myst and what is in sphinx-jupyter.

Eg I could a case for "keep myst jupyter agnostic, it is just a better markdown parser and bridge to sphinx". Then, put the ipynb parser inside of sphinx-jupyter and have it leverage myst. There is nothing inherently jupyter-focused about myst, right? I think the main clash would be that we may not actually want to use sphinx-jupyter to run the notebook cells

chrisjsewell:
As I've mentioned before, I would like to see the `jupyter-execute` directive split into `jupyter-execute` and `jupyter-output`, so that (a) the two concepts are not conflated and (b) you can have greater control over the outputs, even possibly having a `jupyter-output` inline role.

Going from a notebook to a text document, this is fairly easy; when you reach a code cell, you just write both directives (execute, then output) after each other (with options based on the cell metadata).

Going from a text document to a notebook is a bit trickier (but not that much); when you reach an execute block, you need to check if there is an output block directly below it, then both of these are extracted, and the code cell contents and metadata constructed appropriately.

choldgraf:
@chrisjsewell maybe something we can pursue in this repo, and then see if it makes sense to upstream there later on?

chrisjsewell:
[...]

Issue #46, opened 2020-02-15
IPYNB parsing: auto-register scrapbook scraps somehow


choldgraf:
An idea I just had when thinking about how we can reference content from one notebook to another - what if as a part of the Sphinx parser we automatically generated docutils targets from any scraps that are in a notebook during the parsing process. That way references to those targets could already exist and users could reference them with a role, similar to what we do with `:ref:`. I'm imagining something like:

## In a notebook

Somebody writes an analysis that generates a plot they'd like to include elsewhere in their docs. They run `scrapbook.glue('myplot')`

## In a MyST document

Somebody wants to include that "scrap" from the notebook. They only need to do something like

```
{scrap}`myplot`
```
and Sphinx just looks up `myplot` against a list of scraps that it has found across any parsed notebook.

chrisjsewell:
This may well be possible with #47

chrisjsewell:
Having looked at a scrapbook, I don't think it does quite what we want. But actually, using the same kind of mechanics, its actually super easy to achieve this 'content recording'. Basically paste this code into a notebook, execute and save:

```python
def get_mimetypes(obj):
    if hasattr(obj, "_repr_mimebundle_"):
        return obj._repr_mimebundle_()
    mimebundle = {}
    for mimetype, method in (
        ("text/plain", "__str__"),
        ("text/html", "_repr_html_"),
        ("application/json", "_repr_json_"),
        ("image/jpeg", "_repr_jpeg_"),
        ("image/png", "_repr_png_"),
        ("image/svg+xml", "_repr_svg_"),
        ("text/latex", "_repr_latex_"),
    ):
        if hasattr(obj, method):
            mime_content = getattr(obj, method)()
            if mime_content is not None:
                mimebundle[mimetype] = getattr(obj, method)()
    return mimebundle


def record_outputs(obj, key, metadata=None):
    from IPython.display import display

    mimebundle = get_mimetypes(obj)
    if not mimebundle:
        raise ValueError("No mimebundle available")
    metadata = metadata or {}
    metadata.update({"record_key": key})
    display(
        {"recorded/" + k: v for k, v in mimebundle.items()},
        raw=True,
        metadata=metadata,
    )

a = "abc"
record_outputs(a, "mytext")
import pandas as pd
record_outputs(pd.DataFrame([1, 2, 3]), "mytable")
```

If you look at the notebook JSON, you'll see the outputs have all been saved, along with the keys in the metadata:

```json
{
[...]

Issue #27, opened 2020-02-13
How will notebook content be parsed into Sphinx


choldgraf:
Aside from the question of what markdown flavor to support *within text cells* (see ExecutableBookProject/meta#14 and ExecutableBookProject/meta#19 and ExecutableBookProject/meta#18 etc for that), there is the question of what content will actually be read into Sphinx. Here's a place to discuss this.

## A starting point for planning:

* We assume there is always both a text-based and an `ipynb`-based version of any content file.
* The `ipynb` file contains all of the outputs, and it is the thing that gets read in to Sphinx
* A Sphinx parser loops through the notebook cells, and adds a docutils container around the cell that contains some cell metadata (e.g. tags, and the kind of cell)
* For text cells, the Sphinx parser then parses the contents of the cell w/ a markdown parser (maybe an rST parser optionally?)
* For code cells, it adds another container for "input" and parses the content as a code block. It then parses the output selectively based on what kind of output is in the mimetype.

## How Jupyter Book caches / uses input content

A quick example for illustrative purposes:

The only assumption that Jupyter Book makes about the structure of incoming content is that all content is either

* An `ipynb` file
* A file that can be converted to `ipynb` with Jupytext

Building a Jupyter Book is a two-step process, one step is to run `jupyter-book build`. This does:

* For any text content files, convert them to `ipynb` and run them top to bottom, then convert to HTML in the cache folder
* For any `ipynb` files, assume they've already been run and convert them to HTML in the cache folder

Then, when the full book is constructed, it only uses the cached HTML versions that are in the cache folder to stitch together the book.

## Another approach - riff off of what Jupyter Book does?

As I wrote the above, it got me wondering whether we want to force two versions of the content, and instead to use a cacheing system like the one described above. The main difference would be that instead of the cache step creating an HTML file (like in Jupyter Book), it would create an `ipynb` file that Sphinx would know how to parse into its AST when the whole book was built.

PR #47, opened 2020-02-16
WIP: Proposed improvement/alternative to `jupyter-sphinx`

chrisjsewell:
@choldgraf @akhmerov following discussion in #42 I've created a proposal here for how I would forsee the execution working.

To recap, the [Sphinx build phases](https://www.sphinx-doc.org/en/master/extdev/index.html#build-phases) and [Sphinx core events](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events) can be summarised as:

```raw
1. event.config-inited(app,config)
2. event.builder-inited(app)
3. event.env-get-outdated(app, env, added, changed, removed)
4. event.env-before-read-docs(app, env, docnames)

for docname in docnames:
    5.  event.env-purge-doc(app, env, docname)
    if doc changed and not removed:
        6. source-read(app, docname, source)
        7. run parser: text -> docutils.document
        8. apply transforms (by priority): docutils.document -> docutils.document
        9. event.doctree-read(app, doctree)

10. event.env-updated(app, env)
11. event.env-check-consistency(app, env)

for docname in docnames:
    12. apply post-transforms (by priority): docutils.document -> docutils.document
    13. event.doctree-resolved(app, doctree, docname)

14. call builder

15. event.build-finished(app, exception)
```

`jupyter_sphinx.main` contains the extension setup, and utilises this approach:

**In (2) Setup a `JupyterDB`**, that is stored within the doctrees build folder. This is imported from a separate package `jupyter_db` whose sole purpose is to maintain a database of all the kernels code cells, and their outputs. It doesn't know anything about sphinx, hence having it as a separate package. I have written this as an SQL database (with sqlalchemy to provide an OO layer on top), since I think that's best for fast look-up and ansynchronous access. But you could essentially write this in anyway you see fit, as long as it has the same interface (i.e. class methods)

**During (3)** any documents (and their related cells, kernels, outputs) that have been changed/removed are removed from `JupyterDB`.

**During (7)** directives save kernels and code cells to the `JupyterDB` and the `JupyterDB` return primary keys (pk) to those elements, which are saved on the `docutils` elements for later lookup.

**During (10)** the `JupyterDB` is parsed to an entry-point defined function: `run_execution`. This function's job is to act 'in-place' on `JupyterDB` to populate it with outputs.
Again it doesn't *know* anything about sphinx, so could be external to `jupyter-sphinx`.
To decide which kernels need re-running, it can check which kernels / cells no longer have outputs, or could even check the outputs 'last modified time'.

**During (12)** The primary keys are used to access the outputs, and insert them where necessary.


chrisjsewell:
FYI this is what the internal structure of the SQL database is:

SQL_Structure

akhmerov:
This is really cool! After I understand it better, I think it would be great if this could go to jupyter-sphinx, and fix a bunch of issues there!

akhmerov:
@chrisjsewell a few questions:

- In jupyter-sphinx the notebooks themselves are used to serve to the users, isn't that a natural way to store the inputs and outputs, as opposed to a database? At which step would it negatively influence performance?
- Can you explain more about the `JupyterViewNode`: what is its purpose?

choldgraf:
I'll take a look at this as well, it sounds like quite an interesting proposal!

At a first glance, my main concern is with maintainability and using an SQLite database. It may be the best technical solution to this, but we should think about it balanced against the likelihood that potential maintainers will have experience with sqlite. I can get a better feel for this after looking through the implementation code. To me the question is always "will contributing this project be accessible to somebody with only volunteer time, and a PhD's worth of Python experience in a scientific field"

chrisjsewell:
1. So if you look at `JupyterDB`, there is a `get_notebook()` method that will return an `nbformat` notebook. 
2. `JupyterDB` is written to be ‘backend agnostic’, so you should be able to switch out the SQL backend with anything you want. 
3. The ‘natural’ reason to use an SQL backend, is because cells and kernels (and mimebundles) utilise integer primary keys to reference them, that are unique across *all* documents/notebooks (which are not necessarily one and the same). At the ‘post transform’ stage you need to be able to return a cells outputs by it pk (or a synonymous name string). Using a folder of notebooks as a backend, for example, you would have to work out how to cache file reads, so you don’t have to re-read a notebook every time you want to access a cell. You would then have to keep an internal database of pk’s to cells/kernels.

chrisjsewell:
> To me the question is always "will contributing this project be accessible to somebody with only volunteer time, and a PhD's worth of Python experience in a scientific field"

Well I think sqlalchemy makes it relatively easy to work with sql databases, an hides a lot of the complexity. I only learnt it in January. But have a look at the code and see what you think.

chrisjsewell:
> Can you explain more about the JupyterViewNode: what is its purpose?

The idea of this is that there is no inherent constraint that the outputs of a code cell must have one defined visualisation. Along the lines of the classic Model-View pattern for data visualisation, ideally the data and the views would be split. Granted that for simple use cases, the user may want to reduce the syntax, and only use the single `jupyter-execute` directive for both. But to have greater control of the output (e.g. controlling whether to render latex as math), this is a better approach. This also opens up possibilities like having an inline role views of the output, and achieving things like #46

akhmerov:
Thanks for the explanation. I agree that storing all outputs, and referring to them from the `JupyterView` node is much better than what we have jupyter-sphinx right now—at the very least to better support multiple writers.

How would the user control rendering of a single output? By including a separate `.. jupyter-view`? 

I would like to suggest sticking to the constraint that the cell outputs only occur after the cell. This choice was a result of a lengthy discussion when we were working on jupyter-sphinx, and the main motivation was the robustness. When working on documentation, it is usually easy to keep in your memory what happens within a single kernel, and manipulate its outputs explicitly. Allowing to include outputs elsewhere may lead to subtle errors because the effects of the changes become much more nonlocal. Similarly, I propose to make the inline role also execute a small cell, rather than refer to some code. I expect that the inline executable code would mainly yield text or latex output, and would be relatively short. This approach (inline code rather than inline references) is also used by Rmd.

akhmerov:
Also to provide more background on juptyer-sphinx design: a principle I found really useful was to critically examine the amount of configurability we allow. For example, I see that you control that whether errors are tolerated is controlled on a per-kernel basis in your implementation, whereas the jupyter-sphinx only has a global setting or a per-cell setting. This is a result of considering usage scenarios outlined in https://github.com/jupyter/jupyter-sphinx/pull/73.

jupyter_sphinx/main.py
```diff
@@ -0,0 +1,130 @@
...
+            "image/jpeg",
+            "text/latex",
+            "text/plain",
+        ],
+        "env",
+    )
+
+    # Document Elements
+    app.add_directive("jupyter-kernel", JupyterKernel)
+    app.add_directive("jupyter-exec", JupyterExec)
```
akhmerov: Another question: is there a reason why you use `jupyter-exec`, and not `jupyter-execute`? I personally don't have a preference either way, but if possible I'd like to avoid changing conventions. Or is there another package that introduces a `jupyter-exec` directive?

chrisjsewell: No it was absolutely an arbitrary choice, and can be changed to `jupyter-execute` 👍

chrisjsewell:
> Also I'd like provide more background about the jupyter-sphinx design

Thanks 👍 , I will certainly take this into consideration. Actually the only reason I added the `:allow-errors` option, was because I wanted to check errors were being correctly renderer in my test case:  https://github.com/ExecutableBookProject/myst_parser/blob/jupyter/tests/jupyter_sphinx/sourcedirs/basic/contents.rst

chrisjsewell:
> How would the user control rendering of a single output? By including a separate ..` jupyter-view`?

Yeh that's the idea:

```restructuredtext
.. jupyter-exec::
    :timeout: 20
    :show-code:

    print("x")
    a = 1
    a

.. jupyter-view::


```

Although as I mentioned before, for simple use cases this may not be ideal; to have to put a `.. jupyter-view::` after every code block. So maybe there is some basic options for rendering on the `jupyter-execute`, but then if you want to have finer output control, you use a `jupyter-view`. Something to think about...

akhmerov:
> So maybe there is some basic options for rendering on the jupyter-execute, but then if you want to have finer output control, you use a jupyter-view.

jupyter-sphinx now defines some basic options like `:hide-output:`. Wouldn't enabling a more fine-grained control over the representation be a worthy extension of the interface once there's a specific need for it?

chrisjsewell:
> jupyter-sphinx now defines some basic options like `:hide-output:`. Wouldn't enabling a more fine-grained control over the representation be a worthy extension of the interface once there's a specific need for it?

Well you still at least need to make sure that the code/process is structured in such a way that it can be extended like this. Also, to my mind there is already some potential needs for it; adding things like captions and labels to outputs, allowing for complimentary latex/PDF output, etc...

akhmerov:
Ah, but I have nothing against the *view node* or the code structure. Rather I think the *view directive* may be omitted at first—that faces the users and complicates the interface. Instead, I imagine the view node would be inserted during the execution.

akhmerov:
Also: what are output captions and labels? Is that similar to latex figure captions and labels? If that's the case, wouldn't it be better specified outside of the jupyter-specific directives?

chrisjsewell:
> Ah, but I have nothing against the _view node_ or the code structure. Rather I think the _view directive_ may be omitted at first—that faces the users and complicates the interface. Instead, I imagine the view node would be inserted during the execution.

Yeh no problem, I wouldn't be completely against that lol

chrisjsewell:
> Also: what are output captions and labels? Is that similar to latex figure captions and labels? If that's the case, wouldn't it be better specified outside of the jupyter-specific directives?

Well it would be like the `figure` directive: https://github.com/docutils-mirror/docutils/blob/e88c5fb08d5cdfa8b4ac1020dd6f7177778d5990/docutils/parsers/rst/directives/images.py#L101

You mean wrapping the `jupyter-execute`? That would make it very difficult to achieve round-trip conversion between the text representation and the notebook.

akhmerov:
I think a round-trip conversion would be nearly impossible to achieve regardless of that, due to the notebook format limitations. Both v2 and v2.1 component listings in https://github.com/ExecutableBookProject/meta/issues/21 don't include round-tripping. See also discussions in #27. Come to think of this, even following this PR, a text representation with more than one kernel cannot round-trip to a notebook.

chrisjsewell:
Well it should be round-trippable for a basic subset of the text format. Yes you won’t be able to do a 1:1 mapping for multiple kernels, but then generally if your just using them for separate sections you can just split those sections into multiple documents (would be nice in Sphinx if multiple source documents could end up as a single HTML page). 

For single kernel pages particularly, it might be nice (and facilitate this round trip) if you could specify the kernel in the front matter. In RST it would look like this:

```restructuredtext
:kernel_name: name
```

In myst:

```markdown
—-
kernel_name: name
—-
```

akhmerov:
> Well it should be round-trippable for a basic subset of the text format.

Sure, but why would figure captions and labels belong to this basic subset?

My concern here is that this way a directive for code execution also takes responsibility for something unrelated to execution, which is also provided by another directive.

chrisjsewell:
> My concern here is that this way a directive for code execution also takes responsibility for something unrelated to execution,  which is also provided by another directive.

Well surely this was my whole point with the `jupyter-view` directive: at present the `jupyter-execute` directive is **not** only responsible for code execution, it also takes responsibility for the output rendering.

This is not already provided by another directive; the `figure` directive only works if you have an image uri, which is not the way that `jupyter-execute` works.

akhmerov:
@chrisjsewell great point. Since caption itself may include arbitrary rst content, it also seems hard to implement a wrapper that would create figures.

Continuing with the usage of `jupyter-view`: do you envision that it'd have a reference to the `jupyter-execute` that generates the view? Or would it only be correct to include the view immediately after the execute?

Also: how would this interact with thebelab? Since its users aren't the code devs, they will be confused if the input code is separated in the document from the result it produces (also thebelab only inserts the output right after the input). Should the code appear in the figure body after thebelab is activated? What should happen to the original code? Same questions apply to the case when there are multiple references to the execute call, or when the execute and view are separated in the document. Currently in jupyter-sphinx the code only moves when thebelab is activated if initially if it was rendered below the output, and that's not too far.

Further questions: do we expect that the users will need to tweak the rendering of the mime bundle on a case by case basis, or would a global configuration suffice?

jupyter_db/main.py
```diff
@@ -277,3 +292,55 @@ def add_outputs_from_nb(
...
+        fname_format="mimebundle_{pk}{ext}",
+        session: Optional[Session] = None,
+    ) -> str:
+        with self.context_session(
+            session=session, final_commit=False
+        ) as session:  # type: Session
+            mime: OrmMimeBundle = session.query(OrmMimeBundle).filter_by(pk=pk).one()
+            source = mime.source
+            if mime.mimetype == "html":
+                source = dedent(source)
```
akhmerov: Is this always a safe operation?

chrisjsewell: What do you mean safe? It’s not actually executing the source?

akhmerov: Sorry, that's a noob question: is resenting html guaranteed to not change its looks?

choldgraf:
What's the status of this PR? Is it superseded by the existence of `jupyter-cache`?

chrisjsewell:
> What's the status of this PR? Is it superseded by the existence of jupyter-cache?

Eventually it will be, but I'll leave it open for now since it has some different/additional code than is currently in `jupyter-cache`, that I may need at a later stage.

chrisjsewell:
Now being implemented in `jupyter-cache`

PR #50, opened 2020-02-18
Add colon syntax for directive options.

chrisjsewell:
Also add error reporting for YAML block parsing.

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -421,25 +421,88 @@ class DocutilsRenderer(BaseRenderer):
             self.current_node += problematic
 
     def render_directive(self, token):
+        """parse fenced code blocks as directives.
+
+        Such a fenced code block starts with `{directive_name}`,
+        followed by arguments on the same line.
+
+        Directive options are read from a YAML block,
+        if the first content line starts with `---`, e.g.
+
+        ::
+
+            ```{directive_name} arguments
+            ---
+            option1: name
+            option2: |
+                Longer text block
+            ---
+            content...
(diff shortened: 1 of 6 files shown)


myst_parser/docutils_renderer.py
```diff
@@ -421,25 +421,88 @@ def render_role(self, token):
...
-            # TODO check options are an un-nested dict?
+            except (yaml.parser.ParserError, yaml.scanner.ScannerError) as error:
+                msg_node = self.document.reporter.system_message(
+                    3, "Directive options:\n" + str(error), line=token.range[0]
+                )  # 3 is ERROR level
+                msg_node += nodes.literal_block(yaml_block, yaml_block)
+                self.current_node += [msg_node]
+                return
+            # TODO check options are an un-nested dict / json serialize ?
+        elif content.startswith(":"):
```
choldgraf: Shall we lstrip here and in the while loop? I believe @jstac said he'd imagine indenting the : lines

chrisjsewell: NOW you come in and review, after I moved the review restriction and went ahead and merged this PR 😆 Yeh this makes sense, especially in case the directives are nested, probably also need to handle indentations in the `---` YAML blocks.
If you could add an issue for it or make a new PR ta.

choldgraf: HAH I knew you were gonna say that ;-)

I'll make it a PR

chrisjsewell: FYI I’ve added #54, so a PR would close that

PR #60, opened 2020-02-19
[DNM] showing off weird syntax w/ tabs directive

choldgraf:
This is just a dummy example to show off the strange behavior I am experiencing. There are two "test" files added, both should behave the same way, but the MyST one raises the warning:

```
/myst_parser/docs/test/md.md:2: WARNING: Directive 'tab' arguments:
maximum 0 argument(s) allowed, 1 supplied
```

rendered markdown: https://117-240151150-gh.circle-artifacts.com/0/html/test/md.html
rendered rst: https://117-240151150-gh.circle-artifacts.com/0/html/test/rst.html

(the CSS looks weird, I know, but I think that's a theme issue)


choldgraf:
Yeah I agree, it is a very wonky extension. I mostly opened this issue in case there's an underlying problem it is exposing, but if this extension is buggy or edge casey enough I don't think we need to design around it. What do you think?

chrisjsewell:
Oh no, its definitely good to understand any divergent behaviour with the RST parser.

So from d3c527b, and looking at the [artifact here](https://119-240151150-gh.circle-artifacts.com/0/html/test/md.html), it does as I'd expect from the directive code: read the name from the first content line and read the tab content from the 2nd line on.

So with the RST parser then, it appears that, if there is no required/optional arguments (and no allowed options), it starts reading content from just after the directive name:

```
... name:: this is all content
    and this
    and this
```

Wierd! Is that a bug of the RST parser, that `sphinx-tabs` has inadvertently utilised, or an expected behaviour?

choldgraf:
My opinion is that it's a bug, but I have no idea what the devs intended. It seems to break "explicit is better than implicit" as well as "only one obvious way to do it" rules of python ;-)

chrisjsewell:
In the [docutils syntax documentation](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives), there is no mention of this behaviour. But they do use the example:

```
.. note:: This is a paragraph

   - Here is a bullet list.
```

choldgraf:
Ah interesting - see, there I would assume "this is a paragraph" would have a special meaning from the content block.

So I guess the question is: "do we raise this issue w/ rST and treat it as a bug, or do we mimic its behavior and treat it as a feature"

chrisjsewell:
Yeh hmm, checked in https://122-240151150-gh.circle-artifacts.com/0/html/test/rst.html,
and that is the behaviour, and [this is where](https://github.com/docutils-mirror/docutils/blob/e88c5fb08d5cdfa8b4ac1020dd6f7177778d5990/docutils/parsers/rst/states.py#L2147) its implemented in the docutils code. So guess I'll have to put it in, to be compliant with extensions tht might use it 😕

choldgraf:
that's annoying :-/  I'd say it's worth mentioning but this is a docutils thing, yeah? sounds like they don't have the most efficient processes for making changes etc

chrisjsewell:
Fixed by #79

PR #73, opened 2020-02-22
Add `BlockBreak` token (`+++`)

chrisjsewell:
Following from discussion in ExecutableBookProject/sphinx-notebook#12,
I have added the `BlockBreak` AST token, which parses lines beginning with `+++`.

In the docutils renderer, these lines are converted to `comment` nodes, so they won't be rendered in output text, but also with an added class, so they can be identified by potential sphinx transform extensions:

```
+++ a block break
```

goes to Markdown AST:

```yaml
type: Document
footnotes: {}
children:
- content: a block break
  range:
  - 1
  - 1
  raw: +++ a block break
  type: BlockBreak
```

and docutils AST:

```xml
<document source="">
    <comment classes="block_break" xml:space="preserve">
        a block break
```

### DIFF
--- a/docs/using/syntax.md
+++ b/docs/using/syntax.md
@@ -40,14 +40,14 @@ For more information, also see the [CommonMark Spec](https://spec.commonmark.org
   ```
 - **LineComment**: `% this is a comment`. See {ref}`syntax/comments` for more
   information.
-
+- **BlockBreak**: `+++`. See {ref}`syntax/blockbreaks` for more information.
 
 #### CommonMark tokens
 
 - **HTMLBlock**: Any valid HTML (rendered in HTML output only)
 - **BlockCode**: indented text (4 spaces)
 - **Heading**: `# Heading` (levels 1-6)
-- **SetextHeading**: underlined header
+- **SetextHeading**: underlined header (using multiple `=` or `-`)
 - **Quote**: `> this is a quote`
 - **CodeFence**: enclosed in 3 or more backticks with an optional language name. E.g.:
   ````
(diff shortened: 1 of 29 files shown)


chrisjsewell:
Documentation: https://132-240151150-gh.circle-artifacts.com/0/html/using/syntax.html#block-breaks

docs/using/syntax.md
```diff
@@ -434,6 +435,26 @@ Is below, but it won't be parsed into the document.
...
+You may add a block break by putting `+++` at the beginning of a line.
+This constuct's intended use case is for mapping to cell based document formats,
+like [jupyter notebooks](https://jupyter.org/),
+to indicate a new text cell. It will not show up in the rendered text,
+but is stored in the internal document structure for use by developers.
+
+For example, this code:
+
+```
++++ some text
```
choldgraf: The "some text" doesn't make it into the AST in a structured way, right? And if not I don't think we should add it here just to make sure users don't think it's usable in some way

myst_parser/docutils_renderer.py
```diff
@@ -80,7 +80,8 @@ def render_children(self, token):
             self.render(child)
 
     @contextmanager
-    def set_current_node(self, node, append=False):
+    def current_node_context(self, node, append: bool = False):
```
choldgraf: nice 👍

docs/using/syntax.md
chrisjsewell: Yeh it does, have a look at the examples in my comment, where "a block break" is stored. I was thinking you might want to use it to serialize and 'store' text cell metadata, for round-tripping (even if this metadata is not actually used in sphinx), e.g.

```
+++ {"key": "value"}
```

PR #76, opened 2020-02-22
tiny convenience func to encode a dict as a docinfo node

choldgraf:
This wraps a bit of the FrontMatter renderer into a hidden function that could be reused to copy the same logic. I imagine that over time, this could also be where serialization happens etc...it'd be good if, whenever that happens, we can easily pull the same behavior into sphinx-notebook so I think this is an easy way to do so.

Doesn't change any behavior, only cosmetics

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -108,9 +108,6 @@ class DocutilsRenderer(BaseRenderer):
         since `process_doc` just converts them back to text.
 
         """
-        # TODO this data could be used to support default option values for directives
-        docinfo = nodes.docinfo()
-
         try:
             data = yaml.safe_load(token.content) or {}
         except (yaml.parser.ParserError, yaml.scanner.ScannerError) as error:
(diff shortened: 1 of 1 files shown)


myst_parser/docutils_renderer.py
```diff
@@ -908,3 +897,22 @@ def __getattr__(self, name):
             raise NotImplementedError(msg).with_traceback(sys.exc_info()[2])
         msg = "{cls} has no attribute {name}".format(cls=type(self).__name__, name=name)
         raise AttributeError(msg).with_traceback(sys.exc_info()[2])
+
+
+def _dict_to_docinfo(data):
```
chrisjsewell: I'd rename this just `dict_to_docinfo`, since the `_` denotes a private method, but obviously you'll be looking to import this in `sphinx-notebook`.

chrisjsewell: It could also be an `@staticmethod` on the class, but that's not a deal breaker.

choldgraf: @chrisjsewell but if it were a `@staticmethod` then wouldn't it trip over itself when it did all the `self.current_node` stuff?

choldgraf: (also, I renamed it to not be hidden)

chrisjsewell: Oh I meant just the `dict_to_docinfo` func as it is, but I wouldn't worry about it.

PR #92, opened 2020-03-01
fixing xref literal bug

### LINKED ISSUE #89: links that don't resolve breaks sphinx
I came across a weird bug today:

Links that don't resolve properly seem to break Sphinx. Basically, any link that isn't either a correct target to a filename, or a URL, will result in this error:

```
Exception occurred:
  File "/home/choldgraf/anaconda/envs/dev/lib/python3.7/site-packages/sphinx/writers/html5.py", line 779, in unknown_visit
    raise NotImplementedError('Unknown node: ' + node.__class__.__name__)
NotImplementedError: Unknown node: TextElement
```

I reproduced this by adding any of these lines to the docs of myst_parser:

* `[test](test.html)`
* `[test](test.md)`
* `[test](install.md)` (page actually exists but wrong target name)
* `[test](blahblah)` (target doesn't exist)

but the error is *not* triggered with these links:

* `[test](install)`
* `[test](https://test.html)`

choldgraf:
This closes #89 , but doesn't improve the extra things that @chrisjsewell mentioned as "like to have"s for the xref functionality. For that, I've opened up a new issue: https://github.com/ExecutableBookProject/MyST-Parser/issues/91

This adds a test to make sure that cross-refs are correctly in there as literals until they get resolved.

### DIFF
--- a/tests/test_renderers/test_docutils.py
+++ b/tests/test_renderers/test_docutils.py
@@ -316,6 +316,19 @@ def test_link_reference_no_key(renderer):
     )
 
 
+def test_link_reference_no_exist(renderer):
+    renderer.render(Document(["[name](foo)"]))
+    assert renderer.document.pformat() == dedent(
+        """\
+<document source="notset">
+    <paragraph>
+        <pending_xref refdomain="True" refexplicit="True" reftarget="foo" reftype="any" refwarn="True">
+            <literal classes="xref any">
+                name
+    """
+    )
+
+
 def test_block_quotes(renderer):
     renderer.render(
         Document(
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -590,7 +590,7 @@ class SphinxRenderer(DocutilsRenderer):
         if token.title:
             wrap_node["title"] = token.title
         self.current_node.append(wrap_node)
-        text_node = nodes.TextElement("", "", classes=["xref", "any"])
+        text_node = nodes.literal("", "", classes=["xref", "any"])
         wrap_node.append(text_node)
         with self.current_node_context(text_node):
             self.render_children(token)


tests/test_renderers/test_docutils.py
```diff
@@ -316,6 +316,19 @@ def test_link_reference_no_key(renderer):
     )
 
 
+def test_link_reference_no_exist(renderer):
```
chrisjsewell: This actually just replicates: https://github.com/ExecutableBookProject/MyST-Parser/blob/3d4269f4369b2fcbf22930f2a1998cb1ffdc149c/tests/test_renderers/test_docutils.py#L248

So can be removed.

To update the `test_cross_referencing` expected output to `literal` run:

```console
$ pytest -k test_cross_referencing --force-regen
```

choldgraf: interesting - OK I just did that and it did indeed pass...I am a bit mystified at the test infrastructure, though it seems to work nicely haha.

chrisjsewell: Its another on e of my tips from the top 😉 [pytest-regressions](https://pytest-regressions.readthedocs.io) is a game changer for writing certain tests

choldgraf: To that point, I don't see it mentioned in the test_infrastructure page at all...is that page out of date? https://github.com/ExecutableBookProject/MyST-Parser/blob/develop/docs/develop/test_infrastructure.md

chrisjsewell: Nope, its just a pytest plugin that adds some additional 'fixtures' to help write certain types of tests (ones that generate larger output `dict` or `str`). Its not part of the 'core' test infrastructure, and isn't necessary to write a test, but I could add a suggestion to use it.

choldgraf: I think it's worth mentioning since a new developer who looks through the tests will (I assume) see syntax that is specific to that plugin, and could use help in figuring out how to understand or utilize that syntax. I'll open up an issue about it

PR #109, opened 2020-03-07
Improve API documentaion and add Inter-sphinx references

### LINKED ISSUE #19: A way to programmatically generate docs for syntax additions?
I'm loving all of the documentation improvements that we've made on this. It helps me to wrap my head around the tool and think about how it extends commonmark markdown.

One thing I'm concerned with is that over time we forget to clearly document the departures that we've made from CommonMark. Perhaps there is a way that we can programmatically generate a "reference spec" that can be inserted into the docs, so we know there is always a canonical source of truth for what *should* be in the documentation?

chrisjsewell:
Utilising the new documentation for [mistletoe-ebp](https://mistletoe-ebp.readthedocs.io) 😄 

@choldgraf I believe this closes #19

See: https://256-240151150-gh.circle-artifacts.com/0/html/api/index.html

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,1 +1,2 @@
-api/
+_build/
+_api/
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1,0 +1,11 @@
+(api/main)=
+
+# MyST-Parser API
+
+```{toctree}
+:maxdepth: 2
+
+tokens.rst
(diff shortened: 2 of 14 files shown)


choldgraf:
This looks really nice - I think we can clean up the API docs a little bit (some formatting in the docstrings seemed wonky) but this is a great improvemenet

Issue #113, opened 2020-03-09
Link definitions not propogated to nested parser


chrisjsewell:
Currently, link definitions cannot be used inside directives, e.g.

````
```{note}
[reference]
```

[reference]: https://github.com/ExecutableBookProject
````

will not work, but this will:

````
```{note}
[reference]

[reference]: https://github.com/ExecutableBookProject
```

````

This will also be the case for upcoming footnotes

chrisjsewell:
there are two places where a new document is parsed:

- https://github.com/ExecutableBookProject/MyST-Parser/blob/c64f7068ba786a79561adb0017194ebc8bf15ba3/myst_parser/docutils_renderer.py#L108
- https://github.com/ExecutableBookProject/MyST-Parser/blob/c64f7068ba786a79561adb0017194ebc8bf15ba3/myst_parser/docutils_renderer.py#L838

The first fix, is that both should be called with `reset_definitions=False`, to ensure we do not wipe currently collected definitions from the global parse context.

For context, these references are only resolved if the definition is immediately available. To achieve this for a normal document parse in mistletoe, we run a first parse of the source text, where span tokens are not yet processed (stored as raw strings in `SpanContainer`), then a subsequent 'walk' of the syntax tree to process these containers.

chrisjsewell:
@choldgraf something to consider in MyST-NB, currently during the `Document.read` (per cell) definitions are reset; so definitions in one text cell cannot be used by a reference in another cell. This is a different behaviour to if the notebook was e.g. converted to a text document with Jupytext and parsed as a whole.

https://github.com/ExecutableBookProject/MyST-NB/blob/c9478b70daeb34be30c48e807e3aa2ec5611f420/myst_nb/parser.py#L78

choldgraf:
Interesting - how do you imagine it would be better to convert the `ipynb` to a text file (maybe with `{execute}` markdown cells?) and parsing that directly? One downside there is that then `ipynb` *with outputs* wouldn't work.

Is there a way that we could persist information like footnotes etc across cell tokenizing?

chrisjsewell:
I'm not sure of an ideal solution off-hand; just somehow you need to 'defer' the processing of Markdown span tokens until you've found all possible link/footnote definitions are found.

[...]

PR #117, opened 2020-03-11
Add Footnote Syntax

### LINKED ISSUE #98: Add footnote syntax?
As discussed in #31 it appears that people are for a footnote syntax, as implemented in some flavours of Markdown (see [Extended Syntax](https://www.markdownguide.org/extended-syntax/#footnotes)):

```markdown
Here's a simple footnote,[^1] and here's a longer one.[^bignote]

[^1]: This is the first footnote.

[^bignote]: Here's one with multiple paragraphs and code.

    Indent paragraphs to include them in the footnote.
```

Can we come to a decision on this, bearing in mind

1. that this syntax extension is **not** strictly compliant with CommonMark, [as discussed here](https://talk.commonmark.org/t/footnote-extensions-vs-link-reference-definition/2571/9)
2. That (as with any syntax extension) this would incur some parsing performance cost (although probably minimal)

Note these footnotes would be inherently auto-numbered, as can be tried in the VS Code standard preview (GitHub markdown does not support footnotes)

image

paging @choldgraf @mmcky @jstac @akhmerov @najuzilu

chrisjsewell:
This PR introduces footnote syntax to close #98. It also partially addresses #113 and #116 

The actual `Footnote` and `FootnoteReference` markdown tokens are implemented upstream in `mistletoe-ebp`:

- https://mistletoe-ebp.readthedocs.io/en/latest/api/extension_tokens.html#footnote
- https://mistletoe-ebp.readthedocs.io/en/latest/api/extension_tokens.html#footreference

**Important**: As the new documentation notes, this syntax currently only supports single line footnote definitions. Multiline footnote definitions will be addressed in a later PR.

For the docutils rendering, In my test I noted that, for rST, docutils/sphinx actually seems to handle these pretty poorly;

- not auto-collecting them into a single section at the bottom of the page,
- not making sure the numbers are ordered, by order of reference
- not discarding unreferenced footnotes

In the [sphinx doc](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#footnotes), they suggest putting them under a header. But I don't see that being an intuitive approach to footnotes.

Therefore, I have 'fixed' all these issues, as you can see in the test example.

[tests/test_sphinx/sourcedirs/footnotes/footnote_rst.rst](https://github.com/ExecutableBookProject/MyST-Parser/blob/add-footnote/tests/test_sphinx/sourcedirs/footnotes/footnote_rst.rst):

image

[tests/test_sphinx/sourcedirs/footnotes/footnote_md.md](https://github.com/ExecutableBookProject/MyST-Parser/blob/add-footnote/tests/test_sphinx/sourcedirs/footnotes/footnote_md.md):

image

New user documentation of the syntax is here:
https://272-240151150-gh.circle-artifacts.com/0/html/using/syntax.html#footnotes,
and I also improved other aspects of it; turning the syntax lists into tables: https://272-240151150-gh.circle-artifacts.com/0/html/using/syntax.html#block-tokens (I'm sure @choldgraf will appreciate 😄)

@mmcky @rossbar @najuzilu @akhmerov have a look at the above documentation and see what you think!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,20 @@
 [![Documentation Status][rtd-badge]][rtd-link]
 [![Code style: black][black-badge]][black-link]
 [![PyPI][pypi-badge]][pypi-link]
-<!-- [![Conda][conda-badge]][conda-link] -->
+[![Conda][conda-badge]][conda-link]
 
 An extended commonmark compliant parser, with bridges to docutils & sphinx.
 
 ## Usage
+
+To install the MyST parser, run the following in a
+[Conda environment](https://docs.conda.io) (recommended):
+
+```bash
+conda install -c conda-forge myst-parser
+```
+
+or
 
 ```bash
 pip install myst-parser[sphinx]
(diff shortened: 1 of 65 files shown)


chrisjsewell:
The only thing that I now recall could do with improving, is the number ordering of footnote references that are in directives. The issue here is that

1.  For the standard markdown parsel the full document is read, without 'processing' roles/directives. When a 'new' footnote reference is encountered, it is added to a global list, which later provides the numbering order.
2. The roles/directives are now processed in a second parse, and any 'nested parsing' is undertaken. Meaning that any new footnote references encountered here will be added to the end of this global list, rather than in the actual order they occur in the document.

Instead of using this list, I probably need to do a final walk of the docutils syntax tree,  and record the referencing order.

chrisjsewell:
> The only thing that I now recall could do with improving, is the number ordering of footnote references that are in directives

Fixed 👍 

image

chrisjsewell:
rebased for merge in #119

PR #128, opened 2020-04-15
Add paragraph wrapper to table rows

### LINKED ISSUE #127: Issue with urls inside markdown tables
Hi! I'm trying to upgrade [the turing way book](https://the-turing-way.netlify.com/introduction/introduction) to the new [jupyter-book beta version](https://github.com/ExecutableBookProject/cli) and found a bug when I want to link a url inside a table. I'm kind of new in the world of sphinx, commonmark and parsers, so this might be a bug I need to report to sphinx instead of here. I'm sorry if that is the case!

To give a reproducible example of the error, if I want to build jupyter-book using a single markdown file containing the following:
```
| Prerequisite | Importance | Notes |
|---|---|---|
| Version Control | Very Important | |
| Reproducible Environments | Very Important | [Binder](https://mybinder.org/) |

```
(rendered markdown version below)
| Prerequisite | Importance | Notes |
|---|---|---|
| Version Control | Very Important | |
| Reproducible Environments | Very Important | [Binder](https://mybinder.org/) |

I get the following error:

```
Exception occurred:
  File "/Users/martina.gonzales/anaconda3/envs/executable-book/lib/python3.7/site-packages/sphinx/writers/html5.py", line 211, in visit_reference
    assert len(node) == 1 and isinstance(node[0], nodes.image)
AssertionError
```

But if I link it outside the table:

`[Binder](https://mybinder.org/)`

there is no issue.

One possible reason for the issue, but I might be totally wrong in this: Reading the [sphinx error message](https://github.com/sphinx-doc/sphinx/blob/2fac698e764ac28dec86844624f4ac415ea11a37/sphinx/writers/html5.py#L220) in context, it seems that the writer in sphinx is expecting the node parent of the reference to be a TextElement. But when I used the MyST API to manually convert these texts to docutils I get that the parent of the reference is `<entry>`, which I don't think is of text kind. In the isolated reference case (when is not inside the table), the parent node is `<paragraph>`. 

I'm using:
- sphinx [2.4.4] 
- myst-parser [0.8.0]
- python [3.7.7]

cc @choldgraf

martinagvilas:
Closes #127 

I added a paragraph wrapper to `render_table_row`, and modified the existing tests to capture this wrapper. 

Let me know if a new test should be added to this PR, making sure urls inside the table are rendered as references, or something of the sort.

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -534,7 +534,7 @@ class DocutilsRenderer:
                     entry["classes"].append(style)
                 with self.current_node_context(entry, append=True):
                     with self.current_node_context(para, append=True):
-                            self.render_children(child)
+                        self.render_children(child)
 
     def render_math_inline(self, token):
         content = token.content
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -528,11 +528,13 @@ class DocutilsRenderer:
         with self.current_node_context(row, append=True):
             for child in token.children:
                 entry = nodes.entry()
+                para = nodes.paragraph("")
                 style = child.attrGet("style")  # i.e. the alignment when using e.g. :--
                 if style:
                     entry["classes"].append(style)
                 with self.current_node_context(entry, append=True):
-                    self.render_children(child)
+                    with self.current_node_context(para, append=True):
+                            self.render_children(child)
 
     def render_math_inline(self, token):
         content = token.content
--- a/tests/test_renderers/fixtures/sphinx_directives.md
+++ b/tests/test_renderers/fixtures/sphinx_directives.md
@@ -285,15 +285,19 @@ table (`sphinx.directives.patches.RSTTable`):
             <thead>
                 <row>
                     <entry>
-                        a
+                        <paragraph>
+                            a
                     <entry>
-                        b
+                        <paragraph>
+                            b
             <tbody>
                 <row>
                     <entry>
-                        1
+                        <paragraph>
+                            1
                     <entry>
-                        2
+                        <paragraph>
+                            2
 .
 
 --------------------------------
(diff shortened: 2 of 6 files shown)


chrisjsewell:
Thanks @martinagvilas looks good (well apart from the currently failing tests lol)
While you are at it, could I also ask you to add one or two more tests; primarily to test links in table cells.

Cheers, Chris

martinagvilas:
I fixed the failing tests, but I still have to add the new ones

martinagvilas:
I added some tests, let me know if they are enough

Also, I noticed that when using `to_html` with the MyST API it is still rendered without the paragraphs, but if I understand correctly in this case you are using directly the renderer of markdown_it

choldgraf:
@martinagvilas amazing!

yep, the `docutils` writer is different, and IMO it's probably fine to have the outputs of both slightly different, though maybe @chrisjsewell has thoughts if we want to enforce that HTML outputs are equal for both docutils -> sphinx -> html and for the direct "to html" functionality.

chrisjsewell:
Ta, yes HTML is strictly adherent with markdown-it:

- https://github.com/markdown-it/markdown-it/blob/master/test/fixtures/markdown-it/tables.txt
- https://github.com/ExecutableBookProject/markdown-it-py/blob/master/tests/test_port/fixtures/tables.md

martinagvilas:
Great! Just wanted to make sure I wasn't missing anything

chrisjsewell:
btw @choldgraf you can see here that with `on: [push, pull_request]` only runs the tests once for forked PRs

choldgraf:
wooo congrats @martinagvilas on your first pull request merged in the ExecutableBookProject! :-)

@chrisjsewell - thanks for pointing it out - I'll just start forking and making PRs that way from now on

PR #129, opened 2020-04-22
using sphinx book theme docs

choldgraf:
This uses the `sphinx-book-theme` for the documentation. Also adds a github action that previews the CircleCI artifact (which won't work on this PR but will work on subsequent ones)

Demo here: https://350-240151150-gh.circle-artifacts.com/0/html/index.html

For some reason, the glyphs in the admonition blocks aren't showing up in this theme. This seems somehow related to myst_parser, since it works fine in the same PR for myst_nb (https://github.com/ExecutableBookProject/MyST-NB/pull/164)

image

### DIFF
--- a/.github/workflows/artifact.yml
+++ b/.github/workflows/artifact.yml
@@ -1,0 +1,12 @@
+on: [status]
+
+jobs:
+  circleci_artifacts_redirector_job:
+    runs-on: ubuntu-latest
+    name: Run CircleCI artifacts redirector
+    steps:
+      - name: GitHub Action step
+        uses: larsoner/circleci-artifacts-redirector-action@master
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


choldgraf:
Upon further experimentation: if I create a one-page sphinx website with one rst file using the sphinx-book-theme:

* If I enable no extensions the glyphs break
* If I enable myst_nb the glyphs work

So, something is getting enabled in myst-nb that we need to enable in sphinx-book-theme

choldgraf:
Ugh I figured it out - the `font-weight` has to be 900 for these icons or they don't work. FML.

choldgraf:
alright, new SBT release out and bugs fixed https://351-240151150-gh.circle-artifacts.com/0/html/index.html

choldgraf:
OK I think this is ready to go - looks all-ok with the new SBT improvements

PR #130, opened 2020-04-30
trailing whitespace in code fences error

choldgraf:
I noticed that if somebody adds a code fence but puts trailing whitespace in the first fence, then an error results. I couldn't quickly figure out where the change needed to be made, but I added a test that I believe replicates the error (in this PR).


tests/test_renderers/fixtures/syntax_elements.md
```diff
@@ -139,6 +139,18 @@ foo
         foo
 .
 
+--------------------------
+Fenced Code no language with trailing whitespace:
+.
+```
```
choldgraf, line 145: this is the exact same as the test before, but w/ two spaces after the first fence

chrisjsewell:
Possibly after:
https://github.com/executablebooks/MyST-Parser/blob/a084975f02c0b4a9141f75878f78b48afa9f9b5a/myst_parser/docutils_renderer.py#L336

Add `language = language.strip()`

choldgraf:
OK I think that fixed it!

choldgraf:
after I resolved conflicts, I think that the only changes here is just the addition of a test for this bug haha

Issue #140, opened 2020-05-11
Can't add targets to images?


choldgraf:
I just tried adding an image with a `:target:` in MyST, and I got this sphinx error:

```
...jupyter-book/docs/content/content-blocks.md:37: WARNING: Directive 'image' cannot be mocked: MockingError: MockState has not yet implemented attribute 'parse_target'
```

Just want to confirm that this is because MyST-parser doesn't yet have the syntax to parse targets?

And as an aside - how would we recommend people include an image *with a link* that uses MyST markdown instead of `![]()` syntax?

chrisjsewell:
The first thing to do is find where the `parse_target` method is implemented in the docutils (GitHub mirror) source code and link here

choldgraf:
https://github.com/docutils-mirror/docutils/blob/e88c5fb08d5cdfa8b4ac1020dd6f7177778d5990/docutils/parsers/rst/states.py#L1927 ?

chrisjsewell:
Ta, so what we need to now do is implement this method on https://github.com/executablebooks/MyST-Parser/blob/f32bd380bdec755238566fa91352f9ed1d491036/myst_parser/mocking.py#L58

choldgraf:
we want it to be the exact same method, right? There shouldn't be any difference between how myst parses this kind of thing from rST?

chrisjsewell:
From a brief scan, there should be no difference. Although obviously we should add a test or two with the PR to clarify

choldgraf:
See https://github.com/executablebooks/MyST-Parser/pull/141 for a fix (I think)

PR #141, opened 2020-05-11
supporting image targets with the parser

choldgraf:
closes https://github.com/executablebooks/MyST-Parser/issues/140

Note that we had a flake8 error come outta nowhere (I think flake8 was updated?) so I fixed those


choldgraf:
merging this once tests pass since nobody has brought up any issues

Issue #148, opened 2020-05-26
Nested markup in link not working with cross-reference


asmeurer:
If I have a link like 

```
[`page`](something.html#reference)
```

The nested markup in `page` works fine. But if I instead use `reference` as a cross-reference like

```
[`page`](reference)
```

The `page` part isn't rendered with the nested markup.

chrisjsewell:
> The page part isn't rendered with the nested markup.

As in it shows as `` `page` `` instead of `page`?

asmeurer:
Here is what I mean. The link to `ENCODING` should look like it does [here](https://www.asmeurer.com/brown-water-python/docs-myst/helper-functions.html#untokenize-iterable), but it is rendering like [this](https://www.asmeurer.com/brown-water-python/docs-myst/helper-functions.html#untokenize-iterable).

Here is the code https://github.com/asmeurer/brown-water-python/pull/2/commits/c7c62d46ed988c90383ee1a041c1ea747c52100f#diff-34534a52d2a004b4add6ed5bc4cf09f3R10

chrisjsewell:
> Here is what I mean. The link to ENCODING should look like it does here, but it is rendering like this.

These links are the same lol

asmeurer:
Copy-paste error. It should be fixed now.

choldgraf:
So the problem is that "encoding" should have a `pre` tag, but in the myst-parser version it does not (it has a `std std-ref` tag).

So, it should be `ENCODING`, and instead it is rendered as ENCODING.

PR #152, opened 2020-05-31
Fix nested parsing of reference text

chrisjsewell:
Currently, this just outlines the problem, relating to #115 and #148:

In `tests/test_sphinx/sourcedirs/references/index.md` we have references to an external and internal target, with *emphasised* text.
In `tests/test_sphinx/test_sphinx_builds/test_references.xml` you can see that this is correctly encoded in the doctree.
However, in `tests/test_sphinx/test_sphinx_builds/test_references.resolved.xml` you can see that the internal reference now has been stripped of the emphasis.

This is because of sphinx's [ReferencesResolver.resolve_anyref](https://github.com/sphinx-doc/sphinx/blob/c063c9c0fe383a20da613448d1f606b9635773b9/sphinx/transforms/post_transforms/__init__.py#L105), which in turn calls [StandardDomain._resolve_ref_xref](https://github.com/sphinx-doc/sphinx/blob/c063c9c0fe383a20da613448d1f606b9635773b9/sphinx/domains/std.py#L806), at which point `sectname = node.astext()` turns the nested syntax into plain text. I don't see any reason why they needed to have done this, other than just rST can't parse nested inline syntax, so the issue never came up.

To "fix" this then, either an upstream PR needs to be made to sphinx (probably best in the long-term) and/or myst-parser needs to override this behaviour in some way.

cc'ing @choldgraf @asmeurer @rossbar and @akhmerov who have been involved in the related issues


chrisjsewell:
a004754 allows for nested syntax in reference text of markdown style references, e.g. `[some *syntax*](ref)` and also allows for file extensions when referencing document names, e.g. `[](index.md)`

choldgraf:
First off

> also allows for file extensions when referencing document names, e.g. index.md

oh shit that would be amazing

regarding this PR:

I think it looks good to me, and seems really helpful. The one caveat is that we're adding a decent-sized chunk of code to our codebase, so I agree that in the long term this would be great to be upstreamed to Sphinx so that we don't have to care for it (particularly if it's just a matter of patching a few lines in Sphinx rather than adding a whole new xref type).

So I'd be +1 on merging this if others like the PR as well, and then opening an issue about upstreaming it to Sphinx.

chrisjsewell:
> I tested this and it appears to fix the issue.

cool cheers, I'll merge soon, probably wanted to add some more tests first

chrisjsewell:
Ok this is ready to merge @choldgraf @asmeurer at the bottom of https://404-240151150-gh.circle-artifacts.com/0/html/using/syntax.html#targets-and-cross-referencing I added a small addition to the documentation, but let me know if you want more

docs/using/syntax.md
```diff
@@ -757,14 +757,16 @@ Alternatively using the markdown syntax:
...
 ```md
 {any}`my text <header_target>`
 ```
 
-Using the same example, see this ref: [](syntax/targets), and here's a ref back to the top of
-this page: [my text](example_syntax).
+but can also accept "nested" syntax (like bold text) and will recognise document paths that include extensions (e.g. `using/syntax` or `using/syntax.md`)
+
+Using the same example, see this ref: [](syntax/targets), here is a reference back to the top of
+this page: [my text with **nested** $\alpha$ syntax](example_syntax), and here is a reference to another page (`[](intro.md)`): [](intro.md).
```
choldgraf, line 770:
````suggestion

```{note}
If you wish to have the target's title inserted into your text, you can
leave the "text" section of the markdown link empty. For example, this
markdown: `[](using/syntax.md)` will result in: [](using/syntax.md).
```
````

File: myst_parser/__init__.py [hunk 2]
```diff
@@ -4,9 +4,11 @@
 def setup(app):
     """Initialize Sphinx extension."""
     from myst_parser.sphinx_parser import MystParser
+    from myst_parser.myst_refs import MystReferenceReslover
```
choldgraf, line 7: shouldn't this be `Resolver` instead of `Reslover` (and throughout)?

choldgraf:
just two nits in there, let me know when those are resolved and I'm +1 on merge

myst_parser/__init__.py
chrisjsewell: oops yeh fixed

chrisjsewell:
FYI @choldgraf your note addition failed, because the document references are relative (i.e. `syntax.md` not `using/syntax.md`)

choldgraf:
ah oops - I got thrown off by the other examples just before it, but those were just inline literals. OK I'm giving this a merge! Thanks @chrisjsewell, this is an awesome addition!

PR #157, opened 2020-06-03
Removing leading backtick for inline Role snippet.

dhermes:


docs/using/syntax.md
```diff
@@ -179,7 +179,7 @@ sections below:
   - See {ref}`syntax/roles` for more
   information.
   - ```md
-    `{rolename}`interpreted text`
```
dhermes, line 182: I'm just getting going with MyST-Parser (i.e. far from an expert) but I based this change on https://myst-parser.readthedocs.io/en/latest/using/syntax.html#roles-an-in-line-extension-point (and on empirical usage of the feature).

Thanks a lot for this project!

choldgraf:
nice! ah this one has been bugging me for a while now, thanks for fixing it :-)

The error you're getting isn't related to this PR so I'm gonna merge and open an issue to track it. Thanks so much!

PR #165, opened 2020-06-09
adding more getting started docs

choldgraf:
This adds some more getting started information and background on Sphinx / parsers / roles / directives / etc. It tries to address a few more of @amueller's points about assumptions we're making about readers.

The biggest addition is a section describing what directives / roles are, their structure, and a bit of information about what they do. I'm thinking of this content as a complement to what is in `syntax.md`, which is more like a "reference" and complete set of information.

### DIFF
--- a/docs/using/intro.md
+++ b/docs/using/intro.md
@@ -71,13 +71,14 @@ markdown documents. This means that you can use the `.md` extension for your pag
 and write MyST markdown in these pages.
 
 ```{tip}
-MyST markdown is kind-of like two flavors of markdown wrapped in one.
+MyST markdown is a mixture of two flavors of markdown:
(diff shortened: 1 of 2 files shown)


asmeurer:
Wrong @

choldgraf:
Lol goddammit, well Andy I'm welcome to your thoughts as well haha

chrisjsewell:
Well whatever, AM wants to review, it looks like a decent iterative improvement to me

choldgraf:
OK, I just added in more language describing the rST versions of roles/directives etc, so I think all of the issues are addressed! If tests are happy, I'll give this one a merge. Thanks for the feedback everybody :-)

PR #170, opened 2020-06-18
Move sphinx into main requirements

chrisjsewell:
Sphinx was originally set as an extra requirement, when myst-parser contained plugins for the markdown parser. This is no longer the case, so it is now not likely that the package would be used without sphinx, and just overcomplicates things to have to install `myst-parser[sphinx]`, rather than just `myst-parser`.
Note the `sphinx` extra has been left in (as an empty list) so that existing requirements files do not break.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
           keys:
             - cache-pip
       - run: |
-          pip install --user -e .[sphinx,rtd]
+          pip install --user -e .[rtd]
       - save_cache:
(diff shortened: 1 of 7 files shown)


chrisjsewell:
> Note the sphinx extra has been left in (as an empty list) so that existing requirements files do not break.

That being said, we may still want to release this as a new major version (i.e. 0.9.0); in case it does cause any installation issues

choldgraf:
nice - good catch

PR #175, opened 2020-06-23
removing circleci

choldgraf:
I just found out that ReadTheDocs can build documentation for forked pull requests and provide a link to demo it, so we no longer need CircleCI! Yay reduced complexity \o/


choldgraf:
once we merge this I will tell CircleCI to stop building this repo

chrisjsewell:
Interesting, is this a new feature?
Should be fine, but for future knowledge:
Is there documentation on it?
How do you set it up on the repository?
Is there any specific limits to how many builds you can run?

choldgraf:
I think it's relatively new: https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

I'm not sure whether or not there are guarantees about this out there...though if it doesn't work out, it's just a matter of re-enabling CircleCI on these repos

chrisjsewell:
yep cool, go for it 👍

chrisjsewell:
@choldgraf I've just realised an issue here: so with the circleci the documentation was running with the flags: https://github.com/executablebooks/MyST-Parser/blob/ba5c56c391ced762d617c254e93db5e2c808cb1d/docs/Makefile#L23-L24

which means that it runs in "nitpick" mode `-n`, fails on warnings `-W`, but still tries to continue building the documentation even if there is a failure `--keep-going`.
This ensures that the CI fails on warnings, like missing references, but if won't stop at the first warning, so that you can see all the warnings in one run and also view the final documentation.

With this new CI, it currently won't fail on warnings.
You can add https://docs.readthedocs.io/en/latest/config-file/v2.html#sphinx-fail-on-warning, which I guess is equivalent to `-nW`, but there appears to be no option for `--keep-going`, and thus makes it tedious to not be able to identify all warnings in one run.

PR #178, opened 2020-06-26
docs in gha

### LINKED ISSUE #177: Move back to CircleCI(?)
@choldgraf I've just realised an issue here: so with the circleci the documentation was running with the flags: https://github.com/executablebooks/MyST-Parser/blob/ba5c56c391ced762d617c254e93db5e2c808cb1d/docs/Makefile#L23-L24

which means that it runs in "nitpick" mode `-n`, fails on warnings `-W`, but still tries to continue building the documentation even if there is a failure `--keep-going`.
This ensures that the CI fails on warnings, like missing references, but if won't stop at the first warning, so that you can see all the warnings in one run and also view the final documentation.

With this new CI, it currently won't fail on warnings.
You can add https://docs.readthedocs.io/en/latest/config-file/v2.html#sphinx-fail-on-warning, which I guess is equivalent to `-nW`, but there appears to be no option for `--keep-going`, and thus makes it tedious to not be able to identify all warnings in one run.

_Originally posted by @chrisjsewell in https://github.com/executablebooks/MyST-Parser/pull/175#issuecomment-650079264_

choldgraf:
Testing out building docs with GHA

closes #177

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,6 +53,29 @@ jobs:
         file: ./coverage.xml
         fail_ci_if_error: true
 
+
+  docs:
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .[rtd]
+
+      # Build the docs
+      - run:
+        name: Build docs to store
+        command: |
+          cd docs
+          make html-strict
+
+
   publish:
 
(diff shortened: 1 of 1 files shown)


choldgraf:
...not sure why the GHA aren't running...there was a syntax error in the first commit but that should be fixed now

.github/workflows/tests.yml
```diff
@@ -53,6 +53,30 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .[rtd]
+
+      # Build the docs
+      - run:
+        name: Build docs to store
+        command: |
+          cd docs
+          make html-strict
```
chrisjsewell, line 77:
```suggestion
      - name: Build docs to store
        run: |
          cd docs
          make html-strict
```

Issue #179, opened 2020-07-06
Is it possible to make footnotes appear somewhere other than the bottom of the document?


asmeurer:
With RST you can use `rubric: footnotes` to make footnotes appear wherever you want. Is it possible to do this in MyST? It seems to only put them at the bottom of the document.

asmeurer:
Also would https://github.com/executablebooks/MyST-Parser/issues/164 make "rubric: footenotes" do what I want, or would MyST still try to put the footnotes at the bottom?

chrisjsewell:
No, not at present, all footnotes are gathered during parsing, then output at the end of the document in https://github.com/executablebooks/MyST-Parser/blob/877b80cd64cd254abe965b4cfb6b0c7abe2b29ab/myst_parser/docutils_renderer.py#L67

chrisjsewell:
Do you have an example of the kind of outcome you are looking to achieve?

asmeurer:
https://www.asmeurer.com/brown-water-python/tokens.html#op

I generate the table as a rst list table https://github.com/asmeurer/brown-water-python/blob/master/docs/exact_type_table.py. It would be possible, though a little more annoying, to do this with a Markdown table, but I would prefer to keep the footnotes next to the table as they are now.

chrisjsewell:
Yep sounds reasonable; I have done something similar with tables in latex.
Ideally what I think you would want here is a "scoped" footnote section, i.e. at the bottom of your table, you only want footnotes from that table, not footnotes before or after the table

adamgayoso:
+1 for this feature!

Issue #164, opened 2020-06-06
Add an `eval_rst` directive


choldgraf:
In https://github.com/executablebooks/MyST-Parser/issues/163 we identified that there may be circumstances where the author still needs to write rST with MyST. For example, if a person uses `autodoc` and wants their documentation written in MyST, but has docstrings written in rST.

In that case, we could include a directive like `recommonmark`'s `eval_rst` that will call the rST parser on its content.

Another possible approach (maybe this is unrealistic?) is to allow for a flag in *all* directives used by MyST-parser that could tell the directive to parse the content with the rST parser instead of the MyST parser.

chrisjsewell:
> Another possible approach (maybe this is unrealistic?) is to allow for a flag in all directives used by MyST-parser that could tell the directive to parse the content with the rST parser instead of the MyST parser.

From here https://github.com/executablebooks/MyST-Parser/issues/163#issuecomment-640150607, it looks like sphinx does some weird stuff with the `auto` directives, actually rewriting the directive in rST format before parsing. In this case that approach would not work.

choldgraf:
Just a note that it looks like this is where recommonmark does this: https://github.com/readthedocs/recommonmark/blob/c4d575bd7bd5dd992cc3e2bed9dce625f4622a4d/recommonmark/transform.py#L246

It looks like they pass the contents to the RST state parser, which I _think_ we have access to in MyST? We're using a Markdown "mock" parser, so hopefully the rST parser is available as well

asmeurer:
I was thinking, if it isn't hard, it might be useful to also provide the inverse directive, like `.. eval_markdown::` for RST. That way things that currently have to be written in RST like https://github.com/executablebooks/MyST-Parser/issues/163 can still include Markdown segments.

PR #180, opened 2020-07-06
Fix a rendering issue in the API docs

asmeurer:


choldgraf:
nice! thanks for catching this

asmeurer:
I was trying to see if I could figure out #164. I think I see bits and pieces of what needs to be done, but overall it might be simpler if someone who already knows the code fixes it.

choldgraf:
I added a quick comment that points to where I think the relevant code is in recommonmark, maybe that can also help point the right direction?

asmeurer:
Actually I don't need it anymore, so I have less motivation to do it ;)

The only reason I wanted it was #179 and it isn't even clear to me if #164 would fix that.

PR #182, opened 2020-07-09
using rst notes

choldgraf:
Adds some notes about how users can use MyST _alongside_ rST. @asmeurer think this would help get this message across?

here's how it looks: https://myst-parser--182.org.readthedocs.build/en/182/using/intro.html#enable-myst-in-sphinx

### DIFF
--- a/docs/using/intro.md
+++ b/docs/using/intro.md
@@ -46,7 +46,8 @@ Naturally this site is generated with Sphinx and MyST!
 Activating the MyST parser will simply *enable* parsing markdown files with MyST, and the rST
 parser that ships with Sphinx by default will still work the same way. You can have
 combinations of both markdown and rST files in your documentation, and Sphinx will
-choose the right parser based on each file's extension.
+choose the right parser based on each file's extension. Sphinx features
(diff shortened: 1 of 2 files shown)


asmeurer:
You can note that cross references work across the two formats.

docs/using/intro.md
```diff
@@ -40,6 +40,15 @@ To use the MyST parser in Sphinx, simply add: `extensions = ["myst_parser"]` to
 
 Naturally this site is generated with Sphinx and MyST!
 
+```{admonition} You can use both MyST and reStructuredText
+:class: tip
+
+Activating the MyST parser will simply *enable* parsing markdown files with MyST, and the rST
+parser that ships with Sphinx by default will still work the same way. You can have
+combinations of both markdown and rST files in your documentation, and Sphinx will
+choose the right parser based on each file's extension.
```
choldgraf, line 49:
```suggestion
choose the right parser based on each file's extension. Sphinx features
like cross-references will work just fine between the pages.
```

choldgraf:
great - thanks for the suggestion!

PR #217, opened 2020-08-18
👌 IMPROVE: Math Parsing

chrisjsewell:
This centralises parsing configuration storage and validation, meaning it only has to be updated in a single place.
Also moved from using `texmath_plugin` to `dollarmath_plugin`

### DIFF
--- a/docs/using/syntax.md
+++ b/docs/using/syntax.md
@@ -648,7 +648,7 @@ header-rows: 1
 
 (syntax/math)=
 
-### Math shortcuts
+## Math shortcuts
 
 Math is parsed by setting, in the sphinx `conf.py` [configuration file](https://www.sphinx-doc.org/en/master/usage/configuration.html) one or both of:
 
@@ -657,7 +657,7 @@ Math is parsed by setting, in the sphinx `conf.py` [configuration file](https://
 
 These options enable their respective Markdown parser plugins, as detailed in the [markdown-it plugin guide](markdown_it:md/plugins).
 
-#### Dollar delimited math
+### Dollar delimited math
 
 Enabling dollar math will parse the following syntax:
 
@@ -740,7 +740,7 @@ These options can both be useful if you also wish to use `$` as a unit of curren
 
 (syntax/amsmath)=
 
-#### Direct LaTeX Math
+### Direct LaTeX Math
 
 You can enable direct parsing of [amsmath](https://ctan.org/pkg/amsmath) LaTeX equations by setting `myst_amsmath_enable = True` in your sphinx `conf.py`.
 These top-level math environments will then be directly parsed:
@@ -780,7 +780,7 @@ a_{21}& =b_{21}&
 We hope to implement this in a future update (see [executablebooks/MyST-Parser#202](https://github.com/executablebooks/MyST-Parser/issues/202))!
 :::
 
-#### Math in other block elements
+### Math in other block elements
 
 Math will also work when nested in other block elements, like lists or quotes:
 
@@ -806,7 +806,7 @@ Math will also work when nested in other block elements, like lists or quotes:
   a_1=b_1+c_1\\a_2=b_2+c_2-d_2+e_2
   \end{gather*}
 
-#### Mathjax and math parsing
+### Mathjax and math parsing
 
 When building HTML using the [sphinx.ext.mathjax](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#module-sphinx.ext.mathjax) extension (enabled by default), its default configuration is to also search for `$` delimiters and LaTeX environments (see [the tex2jax preprocessor](https://docs.mathjax.org/en/v2.7-latest/options/preprocessors/tex2jax.html#configure-tex2jax)).
 
@@ -829,7 +829,7 @@ To inhibit this override, set `override_mathjax=False`.
 
 (syntax/frontmatter)=
 
-### Front Matter
+## Front Matter
 
 This is a YAML block at the start of the document, as used for example in
 [jekyll](https://jekyllrb.com/docs/front-matter/). Sphinx intercepts these data and
@@ -850,7 +850,7 @@ This is an orphan document, not specified in any toctrees.
 
 (syntax/comments)=
 
-### Comments
+## Comments
 
 You may add comments by putting the `%` character at the beginning of a line. This will
 prevent the line from being parsed into the output document.
@@ -883,7 +883,7 @@ another line
 
 (syntax/blockbreaks)=
 
-### Block Breaks
+## Block Breaks
 
 You may add a block break by putting `+++` at the beginning of a line.
 This constuct's intended use case is for mapping to cell based document formats,
@@ -903,7 +903,7 @@ Is below, but it won't be parsed into the document.
 
 (syntax/targets)=
 
-### Targets and Cross-Referencing
+## Targets and Cross-Referencing
 
 Targets are used to define custom anchors that you can refer to elsewhere in your
 documentation. They generally go before section titles so that you can easily refer
@@ -961,7 +961,7 @@ markdown: `[](syntax.md)` will result in: [](syntax.md).
 
 (syntax/images)=
 
-### Images
+## Images
 
 MyST provides a few different syntaxes for including images in your documentation, as explained below.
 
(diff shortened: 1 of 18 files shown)


chrisjsewell:
Ok @executablebooks/ebpteam and @phaustin I have now hopefully fully solved the Math parsing/rendering.

Basically, I wrote a new markdown plugin in https://github.com/executablebooks/markdown-it-py/commit/91bf225d3ad7162cd08aa61dca1c3389f56c847a, which gives greater control over the parsing process, and fixes the internal escaping issue (as raised in https://github.com/executablebooks/jupyter-book/issues/750#issuecomment-653794950).

Then here, I use that plugin and also re-configure mathjax so that it does not try to replicate this parsing, and so "respects" if the math paring is disabled and makes it consistent with e.g. TeX/PDF outputs.

See https://myst-parser--217.org.readthedocs.build/en/217/using/syntax.html#math-shortcuts, and in particular https://myst-parser--217.org.readthedocs.build/en/217/using/syntax.html#mathjax-and-math-parsing

**Note** because `myst_amsmath_enable=False` by default, this will mean that anyone who currently has "bare" LaTeX in their HTML documentation, which is now being rendered by MathJax, will no longer have it rendered, until they set `myst_amsmath_enable=True`.
We can decide if we want this to be the case, or to set it to `True` by default in myst-nb and/or jupyter-book.

This will then feed into https://github.com/executablebooks/MyST-NB/pull/226 and finally into jupyter-book.

docs/using/syntax.md
```diff
@@ -763,9 +780,56 @@ a_{21}& =b_{21}&
...
+  "displayMath": [["\\[", "\\]"]],
+  "processRefs": False,
+  "processEnvironments": False,
+  }
+}
+```
+
+Since these delimiters are how `sphinx.ext.mathjax` wraps the math content in the built HTML documents.
+
+To inhibit this override, set `override_mathjax=False`.
```
choldgraf, line 828: is there some way for people to provide their own *extra* mathjax overrides? several folks have asked about this in JB

choldgraf:
the docs addition looks good to me, I don't have time right now to take a look at the implementation so perhaps @mmcky can give a review on that. Two quick thoughts:

- Is there a reason not to make "don't parse math if there are dollars signs just before/after" the default? 
- We mention mathjax overrides and that mathjax is the default, but I'm not sure where there's an option to choose anything _other_ than Mathjax. Am I missing something?

Other than those nits I think that this looks really nice!

chrisjsewell:
> Is there a reason not to make "don't parse math if there are dollars signs just before/after" the default?

Not sure what you mean by this? Are you talking about `myst_dmath_allow_space` or can you give an example?

chrisjsewell:
> We mention mathjax overrides and that mathjax is the default, but I'm not sure where there's an option to choose anything other than Mathjax. Am I missing something?

Yes sphinx.ext.imgmath, sphinx.ext.jsmath and https://github.com/hagenw/sphinxcontrib-katex

docs/using/syntax.md
chrisjsewell: Just use `mathjax_config`? Then now, they will also have to ensure `myst_override_mathjax=False`

chrisjsewell:
Right I'm going to merge this, so I can get on to other things, but feel free to have a look through @mmcky and we can always fix anything in post

docs/using/syntax.md
```diff
@@ -721,9 +730,17 @@ $$ (eqn:best)
 
 This is the best equation {eq}`eqn:best`
 
+There are a few other options available to control dollar math parsing:
+
+`myst_dmath_allow_space=False`, will cause inline math to only be parsed if there are no initial / final spaces, e.g. `$a$` but not `$ a$` or `$a $`.
```
mmcky, line 735: @chrisjsewell this is `on by default`?

docs/using/syntax.md
```diff
@@ -721,9 +730,17 @@ $$ (eqn:best)
...
+
+`myst_dmath_allow_digits=False`, will cause inline math to only be parsed if there are no initial / final digits, e.g. `$a$` but not `1$a$` or `$a$2`.
```
mmcky, line 737: @chrisjsewell is this `on by default`?

myst_parser/main.py
```diff
@@ -1,72 +1,118 @@
...
         md.use(container_plugin, "admonition", validate=validate_admonition)
-    if enable_amsmath:
+    if config.amsmath_enable:
         md.use(amsmath_plugin)
+    for name in config.disable_syntax:
+        md.disable(name, True)
+
+    md.options.update(
+        {
+            "enable_html_img": config.html_img_enable,
```
mmcky, line 103: @chrisjsewell is there a reason to have different option names here?

docs/using/syntax.md
chrisjsewell: See the default configuration here: https://myst-parser.readthedocs.io/en/latest/using/intro.html#myst-configuration-options and in the code: https://github.com/executablebooks/MyST-Parser/blob/8206f2a6eb891737b0b7916b761d6a5311a08de4/myst_parser/main.py#L22

I'm open to discussion about these though 😄

myst_parser/main.py
chrisjsewell: Not really, I just did want to make a breaking change to the internal code. But yeh can be changed at some point

mmcky (commented):
@chrisjsewell love your work on this! Many people will really love the features added here. 

In my testing -- the only other edge case I have found is:

## Edge Cases

The following `myst`

```rst
What happens if I use $$
L_x = y
$$
```

produces

Screen Shot 2020-08-19 at 2 57 51 pm

docs/using/syntax.md
chrisjsewell: as in at present the default is that `$ a $` will be parsed as math

mmcky: ah nice page. I just noticed you added `(off by default)` in a couple other places in the text and was wondering if you wanted that here as well?

chrisjsewell: Yeh good point

mmcky: If having a default value of `myst_dmath_allow_space=True` provides a more **robust** (but less technically correct) parse of the text -- I think it is probably good to have on by default. What do you think?

chrisjsewell:
> The following rst

Do you mean that this is what is produced using rst (with a sphinx extension), or that this example was taken from an rst document?

Currently the dollarmath plugin does not support inline double-dollar. It is on the TODO list though

docs/using/syntax.md
chrisjsewell: > provides a more robust (but less technically correct) 

Well I don't know if there is a "technichally correct" syntax rule for this, since obviously its not covered by the CommonMark spec. Just to clarify, you think that `$ a $` should or should not be recognised a math?

mmcky:
> Do you mean that this is what is produced using rst (with a sphinx extension), or that this example was taken from an rst document?

Sorry old habits -- I mean `myst`

> Currently the dollarmath plugin does not support inline double-dollar

OK Roger that. So the display math needs white space on either side of it

These also produce the same output

```md
What happens if I use $$L_x = y$$

What happens if I use
$$
L_x = y
$$
```

but this works:

```md
$$
L_x = y
$$
What happens if I use
```

so leading white space is a hard requirement for `displaymath` -- makes sense

mmcky:
We may want to update this [doc page](https://myst-parser.readthedocs.io/en/latest/using/syntax.html#dollar-delimited-math) as the syntax used shows `$$..$$` on the one line which may confuse people. Want me to add a note?

chrisjsewell:
Yeh pop it in a PR cheers

Basically this is the current specification: https://raw.githubusercontent.com/executablebooks/markdown-it-py/master/tests/test_plugins/fixtures/dollar_math.md

chrisjsewell:
If you (or anyone) feels there are any issues then altered/additional tests should be added to that document (which gets run a parametrized pytests)

mmcky:
> Basically this is the current specification: https://raw.githubusercontent.com/executablebooks/markdown-it-py/master/tests/test_plugins/fixtures/dollar_math.md

So I notice the one line syntax for display math works when next to a `code-block` but not `text`

````
numbered equation following code block. (valid=True)
.
```
code
```
$$a+b$$(1)
.
<pre><code>code
</code></pre>
<section>
<eqn>a+b</eqn>
<span class="eqno">(1)</span>
</section>
````

Do you want me to add a test case

````
numbered equation following text. (valid=False)
.
Immediately after text
$$a+b$$(1)
````

and document `False` for now?

chrisjsewell:
yeh go for it

chrisjsewell:
It wouldn't be too hard to add inline displaymath parsing. However, I would also need to test how sphinx would handle this

Issue #252, opened 2020-09-17
Front matter dates causes the parser to choke


choldgraf:
If you put a date in your YAML front-matter, then the docutils renderer will choke with this error:

```
TypeError: Object of type date is not JSON serializable
```

The error pops up here:

https://github.com/executablebooks/MyST-Parser/blob/master/myst_parser/docutils_renderer.py#L941

I think YAML is parsing the field with an actual Python object, which is then breaking the JSON conversion.

Given that dates are a common field, perhaps we can special-case the date? Or we could add type `date` to the classes that are passed to `str()`

chrisjsewell:
Yeh just add a serialisation rule for datetime objects here: https://github.com/executablebooks/MyST-Parser/blob/59ab5f376c9dd8c65839085a7874b1c12faf2df5/myst_parser/docutils_renderer.py#L940

PR #253, opened 2020-09-17
🐛 FIX: allow dates to be parsed in frontmatter

choldgraf:
This is a small fix to prevent front-matter dates from breaking the docutils parser.

@chrisjsewell could you recommend where a test should go for this?

closes https://github.com/executablebooks/MyST-Parser/issues/252

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -6,6 +6,7 @@ import json
 import os
 import re
 from typing import List
+from datetime import date
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


myst_parser/docutils_renderer.py
```diff
@@ -937,7 +938,7 @@ def dict_to_docinfo(data):
     docinfo = nodes.docinfo()
 
     for key, value in data.items():
-        if not isinstance(value, (str, int, float)):
+        if not isinstance(value, (str, int, float, date)):
```
chrisjsewell, line 941: Also good to add datetime.datetime

choldgraf: done 👍

chrisjsewell (approved):

PR #254, opened 2020-09-19
🔧 MAINTAIN: Use YYYY-MM-DD format in changelog

jpmckinney:


choldgraf (approved):
I'm +1 as long as @chrisjsewell is too as I believe he's the one that wrote these

chrisjsewell:
yep thanks!

PR #256, opened 2020-09-21
🚀 RELEASE: v0.12.10

choldgraf:
This brings in the datetime bugfix...the only other PRs were documentation typo fixes I believe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.12.10 - 2020-09-21
+
+🐛 FIX: allow dates to be parsed in frontmatter.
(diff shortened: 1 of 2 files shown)


choldgraf:
hmmm I got a weird 403 error in the publish to pypi step here: https://github.com/executablebooks/MyST-Parser/runs/1146409808?check_suite_focus=true

let's see if restarting fixes the problem

choldgraf:
That resulted in the same error, here it is:

```
  0%|          | 0.00/42.6k [00:00<?, ?B/s]
100%|██████████| 42.6k/42.6k [00:00<00:00, 220kB/s]
NOTE: Try --verbose to see response content.
HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/
Invalid API Token: InvalidMacaroon('invalid macaroon signature')
```

I'll try again in 30-60 minutes in case this is a pypi bug, and if it still fails I'll delete the release and open an issue

PR #273, opened 2020-12-15
🔀 MERGE: Improve optional syntaxes

chrisjsewell:
Plugins are now split off into `mdit_py_plugins` and more plugins are added:

`myst_enable_extensions=["colon_fence"]` replaces `myst_admonition_enable` and `myst_figure_enable` (which are deprecated) to work exactly like standard back-tick code fences and directives, i.e. you can now use any directive as:

```
:::{name} argument
:option: value

body
:::
```

`myst_enable_extensions = ["dollarmath", ...]` now replaces and deprecates individual enable configuration variables: `admonition_enable`, `figure_enable`, `dmath_enable`, `amsmath_enable`, `deflist_enable`, `html_img_enable`

`myst_enable_extensions=["linkify", "substitution"]` are also big additions

Principally this is all explained in https://myst-parser--273.org.readthedocs.build/en/273/using/syntax-optional.html


chrisjsewell:
@choldgraf, @mmcky, @AakashGfude this adds/changes some key extensions and adds deprecations, so you should review

chrisjsewell:
As explained at https://github.com/executablebooks/MyST-Parser/blob/a21b9bbf42f062ba8f87a8cff7f2eb07b190fc35/myst_parser/docutils_renderer.py#L887:

In would be nice to use these like Jinja substitutions, with filters, e.g. `{{ key | lower }}`,
however, it is not trivial how to handle cyclic substitution referencing with this; you need to somehow identify all the keys used.

Also, error messaging could be improved

chrisjsewell:
Another open question is inline images; within table cells using:

```
---
substitutions:
   key3: "![](img/fish.png)"
---

| |
|-|
|{{key3}}|
```

works, but this does not:

```
---
substitutions:
  key3: |
    ```{image} img/fun-fish.png
    :alt: fishy
    :width: 200px
    :align: center
    ```
```

it just shows the directive as text 🤷

![image](https://user-images.githubusercontent.com/2997570/102515295-786a6a00-4085-11eb-88a4-0edcceb4ca38.png)

docs/using/syntax-optional.md
```diff
@@ -1,32 +1,202 @@
-(syntax-optional)=
+---
+substitutions:
```
choldgraf, line 2: using front-matter YAML for this is pretty cool, IMO

docs/using/syntax-optional.md
```diff
@@ -1,32 +1,202 @@
...
-A special syntax for admonitions can optionally be enabled by setting `myst_admonition_enable = True` in the sphinx `conf.py` [configuration file](https://www.sphinx-doc.org/en/master/usage/configuration.html).
+You can use these substitutions inline or as blocks, and you can even nest substitutions in other substitutions:
 
-The key differences are that, instead of back-ticks, colons are used, and **the content starts as regular Markdown**.
-This has the benefit of allowing the content to be rendered correctly, when you are working in any standard Markdown editor.
-For example:
+:::{tabbed} Markdown Input
 
+```md
+Inline: {{ key1 }}
```
choldgraf, line 115: One concern that I have with using double-curlies `{{` is that it might clash with users that are combining Jinja + MyST Markdown. I could imagine someone, for example, programmatically generating a markdown file with a Jinja template, and then interpolating substitution variables into that file as well. In that case there wouldn't be a way to accommodate both tools, right? I don't know how common this is so not sure if it's worth changing the `{{` (since that is so common for interpolation it makes sense to use it), but wanted to raise this

docs/using/syntax-optional.md
```diff
@@ -1,32 +1,202 @@
...
+
+```
+{{ key1 }}
+```
+
+:::
+
+(syntax/colon_fence)=
+
+## Code fences using colons
```
choldgraf, line 156: I think this makes sense because having the `:::` behave similarly-but-not-quite the same was adding confusion. That said, I think we should make sure to have good "best practice" guides for people - as someone noted in another issue, we're now adding *so much* functionality or flexibility that it can be hard for people to figure out the "right" way to go about doing things.

docs/using/syntax-optional.md
```diff
@@ -1,32 +1,202 @@
...
+
+abc | mnp | xyz
+--- | --- | ---
+123 | 456 | 789
+:::
 ```
 
+````
+
+````{tabbed} Rendered Output
```
choldgraf, line 182: just a note that I guess you wouldn't need to have four ` ```` ` in there if it's only nesting some fences `:::` right? E.g. you could do:

````
```
:::{note}
:::
```
````

File: docs/using/syntax-optional.md [hunk 5]
```diff
@@ -42,21 +212,31 @@ This text is **standard** _Markdown_
...
 
-The supported directives are: admonition, attention, caution, danger, error, important, hint, note, seealso, tip and warning.
-
-These directives do **not** currently allow for parameters to be set, but you can add additional CSS classes to the admonition as comma-delimited arguments after the directive name. Also `admonition` can have a custom title.
-For example:
+and also parameter options can be used:
 
 ```md
-:::{admonition,warning} This *is* also **Markdown**
+:::{admonition} This *is* also **Markdown**
```
choldgraf, line 218: just to be clear - we are now **deprecating** this original `admonition,warning` syntax, right?

docs/using/syntax-optional.md
```diff
@@ -42,21 +212,31 @@ This text is **standard** _Markdown_
...
+
 This text is **standard** _Markdown_
 :::
 
+(syntax/admonitions)=
+
+## Admonition directives
+
+:::{important}
+`myst_admonition_enable` is deprecated and replaced by `myst_enable_extensions = ["colon_fence"]` (see above).
```
choldgraf, line 236: just a note that I think adding a list of extensions to enable is a much better pattern than having a unique variable for each 👍

One extra thought I had - what if we used a single configuration variable like `myst_config` and then had keys / values for our various configurations, rather than using a different variable for each that started with `myst_`?

choldgraf:
I will try and play around with this tomorrow or over the weekend, but took a quick pass and had a few comments in there. In general this is looking great, I like the new config style and love the new features that we're adding in here. Will try to wrap my head around it a bit more.

chrisjsewell:
> Another open question is inline images;

with 08460b5, inline HTML images are now correctly parsed (and can be inserted in tables)

> In would be nice to use these like Jinja substitutions, with filters

implemented in abe0bed!

So now you can do e.g. `{{ env.docname | upper }}` (see updated documentation)

docs/using/syntax-optional.md
chrisjsewell: These substitutions are literally combining Jinja with MyST Markdown now (as of abe0bed), i.e. you can do `{{ key1 | capitalize }}`. So it definitely makes sense that they use `{{` and this approach should be able to replace most of these use cases.
I could easily make the delimiters configurable in the markdown-it plugin, or conversely these delimiters are configurable by the any other Jinja based tool.

docs/using/syntax-optional.md
chrisjsewell: As I explained there, I think this is generally already done, but happy to review if someone wants to try altering the Jupyter book documentation.
In the actual documentation here, I think its pretty clear, whereby there is one file/page for (core) syntax; `docs/using/syntax.md` and one for optional/extended syntax; `docs/using/syntax-optional.md`.
There are then one "core" way an other optional ways, that pepoole can ignore if they want.

The main thing really, is if you care about what the source files look like. If not, just use core, but if you want it to be most readable outside of sphinx (e.g. in notebooks) then use things like `:::` blocks and HTML images

docs/using/syntax-optional.md
chrisjsewell: yep 👍

docs/using/syntax-optional.md
chrisjsewell: yes its deprecated. By deprecated though, I mean entered a deprecation cycle; after this PR, the code will still be parsed and create the same HTML, but a (sphinx) warning will be issued of the deprecation.
Then it will be actually removed in the version after next

This is the same for all the deprecated sphinx configuration options

chrisjsewell: so people should still be able to make their sites/books as before, but they will get a number of warnings, if they were using these features (with messages pointing them towards how to fix)

docs/using/syntax-optional.md
chrisjsewell: meh, I don't think that is necessary or wanted: I believe it would make it difficult in jupyter-book for people to override the defaults of specific aspects of myst-parser, because they would then have to specify the whole dictionary

docs/using/syntax-optional.md
choldgraf: I think that here it's totally fine - the myst parser docs are more "low level" and thus I think should focus on explaining all the features in full over streamlining for a particular workflow. For Jupyter Book perhaps we'll want to re-visit the "introductory" parts a bit more, since that's a more "opinionated" distribution and will probably attract more beginner-level readers that are more likely to get tripped up. I don't think this is actionable on this PR at all though 👍

docs/using/syntax-optional.md
choldgraf: ah interesting - that's pretty neat / powerful. So the use-case I had in mind when I made this comment was something like [the Carpentries](https://carpentries.org/). They have a set of "content templates" that (I think) use Jinja interpolation to generate the actual course material for a specific Carpentry course. The generated material itself might *also* use the interpolation that we describe here (e.g. define a variable in the `conf.py` and re-use it throughout the book, or something like that.

So a couple quick thoughts:

1. I suppose at the Jinja interpolation side, they could interpolate-in variables that had `{{` in the pasted-in content. E.g. they could have the variable `{{ mykey }}` and `mykey` would *equal* `{{ my-other-variable-in-myst-header }}`, that's ultimately what'd get pasted into the MyST Markdown.
2. Alternatively, it sounds like you're saying you could configure Jinja to treat a different delimeter (like `||` or something) to do the interpolation? I didn't realize that was possible!

Sorry if it seems like I am being nit-picky here, I don't think that this is a common use-case, but I am trying to make sure we don't set ourselves up for an anti-pattern here.

docs/using/syntax-optional.md
```diff
@@ -1,32 +1,221 @@
...
+Substitution references are assessed as [Jinja2 expressions](http://jinja.palletsprojects.com) which can use [filters](https://jinja.palletsprojects.com/en/2.11.x/templates/#list-of-builtin-filters), and also contains the [Sphinx Environment](https://www.sphinx-doc.org/en/master/extdev/envapi.html) in the context (as `env`).
+Therefore you can do things like:
+
+```md
+{{ env.docname | upper }}
+{{ "a" + "b" }}
+```
+
+{{ env.docname | upper }}
+{{ "a" + "b" }}
```
choldgraf, line 171: This makes me wonder if there is a way to (in the future, not this PR) incorporate glue-like functionality into this syntax. I think `{{ mykey }}` would be the natural short-hand for ` {glue:}`mykey` `, don't you think? Would using this syntax *both* for "markdown substitutions" **and** for glue-like syntax be possible? (again, I do not think we should do it in this PR, just checking whether it would be possible)

docs/using/syntax-optional.md
choldgraf: makes sense to me - thanks for the explanation 👍

docs/using/syntax-optional.md
choldgraf: makes sense, I think there are pros/cons just from a "readability" perspective

e.g. I think that for a medium-large number of config over-rides:

```
myst:
  extensions:
    - a
    - b
  otheroption: otheykey
```

is more readable than

```
myst_extensions: a
myst_otheroption: b
```

but anyway discussing this I think should be left to the future and need not block this PR

docs/using/syntax-optional.md
chrisjsewell: Oh yeh this was exactly in my thinking

choldgraf: nice - I think using Jinja for this is super cool :-)

docs/using/syntax-optional.md
chrisjsewell: Yeh you can choose whatever you want for the delimiters: https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment

chrisjsewell: `myst_sub_delimiters` now added (and explained in docs)

AakashGfude:
Great work @chrisjsewell . Will read through this, and get familiar with the changes. Sorry, did not get time to review this earlier.
Had a question though, now, that this is merged, I am getting this warning on building docs with pdflatex builder. 
https://github.com/executablebooks/jupyter-book/pull/1125/checks?check_run_id=1581634526 .  What is the resolution to this?

chrisjsewell:
just change the reference to `myst-parser:syntax/optional`

AakashGfude:
> just change the reference to `myst-parser:syntax/optional`

too easy. Thanks.

PR #274, opened 2020-12-18
📚 DOCS: Update API for default parser.

lesteve:
Fix #247.

Took the code from https://github.com/executablebooks/MyST-Parser/issues/247#issuecomment-687251411.

PR #275, opened 2020-12-18
🔧 MAINT: pin sphinxcontrib-bibtex & fix deprecated doc config

lesteve:
Bump into this in #274 so pinning as mentioned in https://github.com/executablebooks/MyST-Parser/pull/274#issuecomment-748017517.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,6 @@ html_title = ""
 html_theme_options = {
     "github_url": "https://github.com/executablebooks/MyST-Parser",
     "repository_url": "https://github.com/executablebooks/MyST-Parser",
-    "expand_sections": ["examples/index"],
 }
 
 # Add any paths that contain custom static files (such as style sheets) here,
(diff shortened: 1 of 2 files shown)


lesteve:
The error was a warning about expand_sections which seems to have been removed from sphinx-book-theme: https://github.com/executablebooks/jupyter-book/pull/1073

choldgraf:
@lesteve perfect -- thanks for the pin and for catching the issue with sphinx-book-theme :-) I suppose we should probably be following our own deprecation patterns shouldn't we ;-)

This looks great to me, thanks very much!

lesteve:
> I suppose we should probably be following our own deprecation patterns shouldn't we ;-)

I guess having multiple repos makes it more likely that one will slip through the cracks :wink:

choldgraf:
you have no idea haha

choldgraf:
actually, you probably do :-)

Issue #284, opened 2021-01-07
Direct support for `meta` directive


tfiers:
[Sphinx's `meta` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-metadata) adds `<meta>` tags to the `<head>` of the page.

This is very useful. One of my usecases is adding `<link rel="canonical">` tags to [Hypothesis-annotated Jupyterbook pages](https://jupyterbook.org/interactive/comments/hypothesis.html) so that the annotations don't disappear when the filename/location changes.

The `meta` directive doesn't work yet alas:
`Directive 'meta' cannot be mocked: MockingError: MockState has not yet implemented attribute 'nested_list_parse'`

I'm up for contributing a patch. This would be my first time contributing to an executablebooks project. Could a maintainer give me a pointer for getting started in the codebase here on this problem?

(my current solution is a custom build script that edits the generated html with `lxml` after `jb build`)

choldgraf:
Agreed this would be useful! I think there are two ways this could be done:

1. Add the method in the error above to `MockState`. I think that'd be a change to the following link w/ a similar pattern to the other methods seen there: https://github.com/executablebooks/MyST-Parser/blob/750ea18db6e7eccc147e3bf45921ac4eb6715ab5/myst_parser/mocking.py#L66
2. Special-case "HTML meta tags" in the FrontMatter for MyST documents (https://myst-parser.readthedocs.io/en/latest/using/syntax.html#front-matter)

number 2 feels a bit more natural to me, though is technically a departure from how rST handles this. Maybe @chrisjsewell has an opinion?

chrisjsewell:
yeh its a bit of a tricky one to mock, in the directive logic.
I'd not firstly though that it is supported, via the use of `eval-rst` (see https://myst-parser.readthedocs.io/en/latest/using/syntax.html#how-directives-parse-content)

````
```{eval-rst}
.. meta::
  :key: value
```
````

As a quick fix, the error message could perhaps mention using this

**Comment by chrisjsewell** [comment]
To elaborate a little further, in:

```
.. meta::
   :description lang=en: An amusing story
   :description lang=fr: Une histoire amusante
```

those `description` are not actually options of the directive, they are its content, but MyST-Parser does not (as yet) have a syntax for field lists.
So even if you implemented `nested_list_parse` it still would not work

chrisjsewell:
but yeh could consider (2)

tfiers:
[...]

PR #285, opened 2021-01-08
👌 IMPROVE: Error msg for unimplemented directives

tfiers:
Suggest using `{eval-rst}` and link to its documentation.

As suggested in https://github.com/executablebooks/MyST-Parser/issues/284#issuecomment-756434092

### DIFF
--- a/myst_parser/mocking.py
+++ b/myst_parser/mocking.py
@@ -243,6 +243,8 @@ class MockState:
                 "https://myst-parser.readthedocs.io/en/latest/using/syntax.html#how-directives-parse-content"  # noqa: E501
             )
         else:
+            # The requested `name` is not a docutils Body element
+            # (such as "footnote", "block_quote", "paragraph", …)
             msg = f"{cls} has no attribute '{name}'"
         raise MockingError(msg).with_traceback(sys.exc_info()[2])
 
--- a/myst_parser/mocking.py
+++ b/myst_parser/mocking.py
@@ -240,7 +240,7 @@ class MockState:
             msg = (
                 f"{cls} has not yet implemented attribute '{name}'. "
                 "You can parse RST directly via the `{eval-rst}` directive: "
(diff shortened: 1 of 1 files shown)


myst_parser/mocking.py
```diff
@@ -235,12 +235,15 @@ def __getattr__(self, name):
...
-            msg = "{cls} has not yet implemented attribute '{name}'".format(
-                cls=type(self).__name__, name=name
+            msg = (
+                f"{cls} has not yet implemented attribute '{name}'. "
+                "You can parse RST directly via the `{eval-rst}` directive: "
+                "https://myst-parser.readthedocs.io/en/latest/using/syntax.html#how-directives-parse-content"  # noqa: E501
             )
-            raise MockingError(msg).with_traceback(sys.exc_info()[2])
-        msg = "{cls} has no attribute {name}".format(cls=type(self).__name__, name=name)
+        else:
```
choldgraf, line 245: Can you add a comment that describes under what condition the "if" statement will return False and thus the simpler error message?

tfiers: Something like this:
```py
if hasattr(Body, name):
    ...
else:
    # The requested `name` is not a docutils Body element
    # (such as "admoniton", "block_quote", "paragraph", …)
    ...
```
?

choldgraf: that seems good 👍

tfiers: I made the code a bit DRY-er but did not introduce the if-else logic, that was already there (i.e. this is a pure refactor -- no different behavior).
Maybe the original author (Chris Sewell) can explain the reasoning behind the two cases.

tfiers: Anyway I added the above comment in a commit, so should be good to go :)

choldgraf:
I believe this is just changing the error message so I think the codecov fail isn't relevant here. Gonna merge away. Thanks so much @tfiers for the improvement :-)

PR #288, opened 2021-01-16
✨ NEW: Add `html_admonition` extension

chrisjsewell:
See https://myst-parser--288.org.readthedocs.build/en/288/using/syntax-optional.html#html-admonitions


choldgraf:
Is this meant to be a solution for people that want to use admonitions styling, but don't have a renderer that supports myst markdown?

chrisjsewell:
yep any thoughts/comments before I merge?

choldgraf:
I think this enhancement sounds pretty good to me...the main concern I have is down a rabbit hole of turning special-cased HTML into "supported" syntax. The medium- and long-term solution here is to get improved support for MyST Markdown into other renderers and UIs, as opposed to bringing HTML syntax directly into MyST. However this "admonition" use-case is common-enough right now that I think it's reasonable to support as a path towards increasing short-term adoption.

Do you agree with that general assessment @chrisjsewell ?

chrisjsewell:
> Do you agree with that general assessment

Yeh, I don't intend to add any more HTML syntaxes, but admonitions and (size controlled) images is a common request (particularly for Jupyter Notebooks) not supported in CommonMark.
I don't see widespread Javascript rendering support happening in the near future, so this is the best path of least resistence.

mmcky:
@chrisjsewell does this pass the contents through as `raw` html or does it get parsed into standard `admonitions` in the `syntax tree`? Ah I think I just found the answer

> These blocks will be converted internally to Sphinx admonition directives, and so will work correctly for all output formats.

this is great 👍 

@AakashGfude support for this should come for free on the `latex` side as long as `admonitions` are supported in general.

PR #301, opened 2021-02-05
👌 IMPROVE: Store rawtext in AST nodes (for gettext)

jpmckinney:
If you have Markdown like:

```md
* My **Awesome** markdown
```

Sphinx needs to be passed `My **Awesome** markdown`.

The code does not do that.

This is the same bug as in recommonmark: https://github.com/readthedocs/recommonmark/pull/187 which had originally been fixed by a Sphinx maintainer, and which I more recently fixed (though recommonmark is very behind on merging PRs).

As you can see in the code for paragraphs, sometimes the original Markdown content is not where you expect it. So far, I've only tested a few tokens.

cc @choldgraf

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -325,7 +325,7 @@ class DocutilsRenderer:
     # ### render methods for commonmark tokens
 
     def render_paragraph_open(self, token: NestedTokens):
-        para = nodes.paragraph(token.children[0].content)
+        para = nodes.paragraph(token.children[0].content if token.children else "")
         self.add_line_and_source_path(para, token)
         with self.current_node_context(para, append=True):
             self.render_children(token)
@@ -449,7 +449,7 @@ class DocutilsRenderer:
             if self.is_section_level(level, self.current_node):
                 self.current_node = cast(nodes.Element, self.current_node.parent)
 
-        title_node = nodes.title(token.children[0].content)
+        title_node = nodes.title(token.children[0].content if token.children else "")
         self.add_line_and_source_path(title_node, token)
 
         new_section = nodes.section()
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -719,7 +719,7 @@ class DocutilsRenderer:
         with self.current_node_context(row, append=True):
             for child in token.children or []:
                 entry = nodes.entry()
-                para = nodes.paragraph("")
+                para = nodes.paragraph(child.children[0].content if child.children else "")
                 style = child.attrGet("style")  # i.e. the alignment when using e.g. :--
                 if style:
                     entry["classes"].append(style)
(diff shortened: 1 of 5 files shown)


jpmckinney:
Actually, the problem is worse. I'll open an issue.

jpmckinney:
Sorry, nevermind - this branch actually does address #304 (at least for the node types I tested).

chrisjsewell:
> Sorry, nevermind 

lol yeh I was just going to say it does capture the "bold" syntax etc.

you can use https://markdown-it.github.io/ (and the `debug` tab) to see what gets stored where in the tokens

myst_parser/docutils_renderer.py
```diff
@@ -325,7 +325,7 @@ def renderInlineAsText(self, tokens: List[Token]) -> str:
     # ### render methods for commonmark tokens
 
     def render_paragraph_open(self, token: NestedTokens):
-        para = nodes.paragraph("")
+        para = nodes.paragraph(token.children[0].content)
```
chrisjsewell, line 328:
```suggestion
        para = nodes.paragraph(token.children[0].content if token.children else "")
```

just to be safe

chrisjsewell (changes requested):

myst_parser/docutils_renderer.py
jpmckinney: Thanks, I added the same condition for headings.

chrisjsewell:
Is there any other additions you want to make before merging?

chrisjsewell:
I see you removed all the ones for list items, etc. Was that intended?

jpmckinney:
@chrisjsewell I added a bunch more test cases in https://github.com/jpmckinney/myst-parser-tests and they all extract correctly now with only the changes to these two lines. I can still add a table test case, and maybe some other block-level elements I misssed. I don't really know how to test this without looking in Sphinx's code to see what it does when the gettext builder is run.

jpmckinney:
I've now added a table test case, and it extracts incorrectly, so I'll add another commit.

jpmckinney:
The new commit fixes terms (`dt`) and table cells.

chrisjsewell:
I think we should add a test for this, to https://github.com/executablebooks/MyST-Parser/tree/master/tests/test_sphinx/sourcedirs:

- Add a `gettext` folder with an index.md and conf.py similar to the ones in https://github.com/jpmckinney/myst-parser-tests
- Add a test to https://github.com/executablebooks/MyST-Parser/blob/master/tests/test_sphinx/test_sphinx_builds.py, that builds that folder with the `gettext` builder and then reads+checks the `_build/gettext/index.pot` is as expected in some way (probably similar to https://github.com/executablebooks/MyST-Parser/blob/702a982b93aa88471b1c40b0d36f344aa5e8e172/tests/test_sphinx/conftest.py#L58-L59)

chrisjsewell:
thats great thanks

chrisjsewell (approved):

Issue #262, opened 2020-10-06
include directive: start from alternative heading level


comabrewer:
Thanks for the great project. I perfectly fits the use case I have, with one minor issue.

**TL;DR**: Is it possible to fix or ignore non-consecutive header levels?

For a group of related projects, I want the contributors to write all the documentation in a single README.md.
From the README and the docstrings, I generate the online docs with Sphinx and MyST.
I want the sections of the README to end up in separate documents of my generated docs.
The `docs/` directory is populated with generic stub files that are generated by a cookiecutter.

The README always has the same sections (second level):
```md
# project-name
## Usage
## Installation
## Changelog
### 0.1.0 (2020-10-01)
## Contributing
...
```

From this, I generate separate pages (e.g. changelog.md):
``````md
# Changelog

```{include} ..\README.md
:start-after: "## Changelog"
:end-before: "## Contributing"
``````

My index.md:

``````md
# project-name

```{toctree}
installation.md
usage.md
api.md
changelog.md
contribution.md
```
``````

What was a level 2 heading now should be a level 1 heading in the split documents.
When building the docs, MyST emits the warning `Non-consecutive header level increase` because of this. The warning is correct and intended, see #71, but is there any way around this? I want to treat Sphinx warnings as errors, because it has proven very helpful in the past.

Possible solutions:

- Is it possible to disable the specific warning? Sphinx offers [suppress_warnings](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings), but I couldn't figure out the type for this warning.
- Is there a possibility to modify header levels of the included documents, or to set the initial level from 0 to 1?

**Comment by chrisjsewell** [comment]
thanks, I'll have to think about this one, perhaps as part of the `include` directive there could be an option to set the initial header level

**Comment by choldgraf** [comment]
Maybe as a first step we can figure out what warning class is raised here so that you could use the suppress warnings config that you mentioned. Surely somebody else in the Sphinx world has had the same problem before? Maybe there’s a stackoverflow question about this or something?

**Comment by chrisjsewell** [comment]
No this warning arises directly from MyST-Parser: https://github.com/executablebooks/MyST-Parser/blob/4e02629f5fb575fdd733ca43891bebb2e312ad00/myst_parser/docutils_renderer.py#L258-L266

correctly, because for example

```md
# title
## sub-title
```

```md
# title
### sub-title
```

```md
## title
### sub-title
```

all are parsed as:

```xml
<section>
[...]

PR #303, opened 2021-02-05
Don't replace documentation content with a warning message

### LINKED ISSUE #302: Issues with sphinx `Locale` transform (assumes rST)
Screen Shot 2021-02-05 at 5 11 55 PM

Documentation repository: https://github.com/open-contracting/standard/pull/1197

To reproduce:

```
pip install -r requirements.txt
make
```

jpmckinney:
closes #302 

I don't know why this library replaces documentation text with warnings. This seems like a bad idea. Warnings should only be output in the console. Anyway, I just replaced the one instance that was breaking my build. I don't know why admonition titles are given a level of "2", but I'll let the maintainers figure that out.

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -285,13 +285,11 @@ class DocutilsRenderer:
         )
 
         if (level > parent_level) and (parent_level + 1 != level):
-            self.current_node.append(
-                self.reporter.warning(
-                    "Non-consecutive header level increase; {} to {}".format(
-                        parent_level, level
-                    ),
-                    line=section.line,
-                )
+            self.reporter.warning(
+                "Non-consecutive header level increase; {} to {}".format(
+                    parent_level, level
+                ),
+                line=section.line,
             )
 
         parent = self._level_to_elem[parent_level]


choldgraf:
huh - yeah that definitely feels like a bug to me 🤔 , I'll leave this for a sec to see whether anybody else had a reason for doing this.

chrisjsewell:
This is not the correct fix for #302

> I don't know why this library replaces documentation text with warnings. This seems like a bad idea. Warnings should only be output in the console.

I think you are misunderstanding this code. Reporter warnings create a [`system_message` node](https://github.com/chrisjsewell/docutils/blob/8adab0660b2097b4f3c32cef7e5ff4cb3c72b084/docutils/docutils/nodes.py#L1772), which is correctly injected into the document AST. This does not replace any documentation text


> I don't know why admonition titles are given a level of "2", but I'll let the maintainers figure that out.

This is not to do with the admonition title; you have a header in your documentation that does not start at 1 (this may be related to https://github.com/executablebooks/MyST-Parser/issues/262).

Why it ends up as a title, I'm not sure, possibly a faulty `Transform` in sphinx

PR #316, opened 2021-02-15
🔧 MAINTAIN: Increase sphinx lower pinning to `v2.1`

jpmckinney:
The EventManager initializer changed signature between 2.0 and 2.1.

MyST-Parser uses the >= 2.1 signature.

https://github.com/sphinx-doc/sphinx/blob/v2.0.1/sphinx/events.py#L44-L45
https://github.com/sphinx-doc/sphinx/blob/v2.1.0/sphinx/events.py#L50-L53


chrisjsewell (approved):
cheers!

PR #335, opened 2021-03-23
🔧 MAINTAIN: Stricter mypy config

hukkinj1:
Mainly adds `check_untyped_defs = true` so that mypy does some basic type analysis even when type annotations are not present.


chrisjsewell:
cheers

PR #344, opened 2021-04-09
🐛 FIX: docutils `v0.17` compatibility

chrisjsewell:
This adds the missing setting that has been added to the RST parser in docutils v0.17.
Although, perhaps it is better to not have `settings_spec` copy/pasted here, but just inherit it

### DIFF
--- a/myst_parser/sphinx_parser.py
+++ b/myst_parser/sphinx_parser.py
@@ -1,8 +1,9 @@
 import time
 from os import path
 
-from docutils import frontend, nodes
+from docutils import nodes
 from docutils.core import publish_doctree
+from docutils.parsers.rst import Parser as RstParser
 from markdown_it.token import Token
 from markdown_it.utils import AttrDict
 from sphinx.application import Sphinx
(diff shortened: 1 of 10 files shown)


myst_parser/sphinx_parser.py
```diff
@@ -116,6 +116,16 @@ class MystParser(Parser):
                 ["--raw-enabled"],
                 {"action": "store_true"},
             ),
+            (
+                'Maximal number of characters in an input line. Default 10 000.',
```
chrisjsewell, line 120:
```suggestion
                "Maximal number of characters in an input line. Default 10 000.",
```

chrisjsewell:
The tests need updating for the HTML outputs: `<div class="section">` -> `<section>`

consideRatio:
@chrisjsewell checking out this PR branch and using docutils 0.17, I could no longer reproduce the error!

chrisjsewell:
> checking out this PR branch and using docutils 0.17, I could no longer reproduce the error!

Thanks @consideRatio, I've changed it again (to just inherit the options directly from docutils) if you want to check again before I merge, and then I'll make a patch release

consideRatio:
Tested to work like this as well! Nice! :)

PR #350, opened 2021-04-21
🔧 MAINTAIN: Remove empty "sphinx" extra

hukkin:
The comment states that this is "left in for back-compatibility", but pip doesn't fail (only warns) if installing nonexistent extras, so I believe this can be safely removed.


chrisjsewell:
cheeers

PR #353, opened 2021-04-24
🔧 MAINTAIN: Move setup to setup.cfg

chrisjsewell:
Also update pre-commits,
fix for docutils 0.17.1,
update sphinxcontrib-bibtex for docs,
add python 3.9 to test matrix.


choldgraf:
Should we do this across all of our repos? And any particular reason you are doing it now?

chrisjsewell:
> Should we do this across all of our repos? any particular reason you are doing it now?

Since `attr` was properly supported for the `version` key: https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v4640

I've already been changing packages to this, but anyhow it will change again when setuptools supports https://www.python.org/dev/peps/pep-0621/ and everything can be moved to `pyproject.toml`

choldgraf:
Yeah i saw that pep too - sounds like projects are in the process of migration now. Sounds good 👍

PR #360, opened 2021-04-28
Add failing test for duplicate labels after translation

jpmckinney:


jpmckinney:
Mysteriously, the anchors_plugin receives the source string (English text) for the regular `#` heading, but the translated string for the admonition's title.

This might help a person more knowledgeable with the code to debug the issue.

PR #362, opened 2021-04-29
⬆️  UPGRADE: Change Sphinx pinning to allow version 4

astrojuanlu:
Sphinx 4.0.0b2 beta is out, and it would be great to see if MyST-Parser works with it. I am doing some local tests with https://github.com/poliastro/poliastro/, which uses it.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,12 +25,15 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        sphinx: [">=3,<4"]
+        sphinx: [">=3,<5"]
         os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


chrisjsewell:
thanks, but I can say for definite there are things to address like https://github.com/executablebooks/meta/issues/306#issuecomment-827681278

astrojuanlu:
Woops, I didn't see that discussion, will take a look. Thanks @chrisjsewell!

jpmckinney:
Can we unpin and later fix mathjax issues? Not everyone who uses MyST-Parser (but who wants to use Sphinx 4) uses mathjax. It can be a minor version to flag that the upgrade isn't seamless.

chrisjsewell:
superceded by #390

PR #363, opened 2021-04-29
📚 DOCS: Add link to `sphinx-external-toc`

hukkinj1:


chrisjsewell:
cheers; I was anyhow thinking of changing the docs a bit to have top-level "pointers":

- I have some existing markdown files that I want to turn into documentation (no/little knowledge of sphinx)
- I have existing sphinx RST documentation that I want to convert / I want to create new documentation utilising docutils/sphinx features

But this can be a stop-gap

PR #369, opened 2021-05-04
✨ NEW: Add `dmath_double_inline` configuration option

chrisjsewell:
Allow display math (i.e. `$$`) within an inline context.
Exposes functionality added in: https://github.com/executablebooks/mdit-py-plugins/pull/19

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -761,7 +761,11 @@ class DocutilsRenderer(RendererProtocol):
 
     def render_math_inline(self, token: SyntaxTreeNode) -> None:
         content = token.content
-        node = nodes.math(content, content)
+        if token.markup == "$$":
+            # available when dmath_double_inline is True
+            node = nodes.math_block(content, content, nowrap=False, number=None)
+        else:
+            node = nodes.math(content, content)
         self.add_line_and_source_path(node, token)
         self.current_node.append(node)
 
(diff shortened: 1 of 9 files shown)


myst_parser/docutils_renderer.py
```diff
@@ -761,7 +761,11 @@ def render_table_row(self, token: SyntaxTreeNode) -> None:
 
     def render_math_inline(self, token: SyntaxTreeNode) -> None:
         content = token.content
-        node = nodes.math(content, content)
+        if token.markup == "$$":
+            # available when dmath_double_inline is True
+            node = nodes.math_block(content, content, nowrap=False, number=None)
+        else:
+            node = nodes.math(content, content)
```
hukkinj1, line 768: @chrisjsewell does this mean that double dollar inline math is rendered differently from single dollar math?

If yes, then I'd argue it's a third distinct type of dollar math syntax, and deserves to be documented as such.

This is important e.g. for formatters which are should then not be allowed to convert double dollars to single dollars or vice versa.

chrisjsewell: >  for formatters which should then not be allowed to convert double dollars to single dollars or vice versa.

I'm a bit "confused" why this would ever be a good idea?
But yeh they render differently, so just use what's in the `Token.markup`

For more context: https://github.com/executablebooks/jupyter-book/issues/1181

hukkinj1: Ah thanks for clarifying, I thought inline `$$` was just an alias for `$` but if I understand correctly it is supposed to be like the block level math block, but the markup is allowed inline.

I still think _technically_ (FWIW, lol), this is a third type of dollar math syntax. No other block level syntax can have their markup inline, and e.g. have the property of not splitting a paragraph in half.

PR #376, opened 2021-05-10
🐛 FIX: Token stream mutation

hukkin:
- Remove two TODOs
- Fix input token stream mutation in `docutils_renderer`. Not sure if myst-parser considers this a bug, though? Markdown-it definitely does, so thought I'd get rid of the mutation.


chrisjsewell:
cheers 👍

Issue #378, opened 2021-05-10
Sphinx 4.0 Support (including Mathjax 3)


chrisjsewell:
Sphinx 4.0 is now out 🎉  and I'm sure there are some that are anxious to upgrade their documentation to use it.

We also certainly want to allow this.
However, there is currently one known blocker for trivially changing the dependency pinning in `setup.cfg`: sphinx-doc/sphinx#7425 / sphinx-doc/sphinx#9094  (Mathjax 3 support) and its interplay with the "dollarmath"/"amsmath" extension.

In https://myst-parser.readthedocs.io/en/latest/using/syntax.html#mathjax-and-math-parsing, I discuss why we want to override default mathjax configuration;
in short because we do not want Mathjax to act as a parser (searching for math, since the Markdown parser already does this), only a renderer (rendering already identified math).

For example, if you set `myst_update_mathjax = False`, and **did not** have "dollarmath" nor "amsmath" in `myst_enable_extensions`, then all of these would still be rendered as math (only with HTML builders but not with other builders like LaTeX)

````
```{math}
a = 1
```

$$b=2$$

\begin{equation}c=3\end{equation}
````

or, another example, if you did have "dollarmath" enabled, but `myst_update_mathjax = False`, then escaped dollar signs would still be rendered as math (again only for HTML builders): 

```
\$\$b=2$$
```

The way we stop this occuring is to override some of the mathjax extension logic in myst_parser/mathjax.py. But this needs updating for Mathjax 3, plus I also want to revisit the logic and check that this is the best approach.

I should be able to do this by next week though 👍 

Cheers,
Chris

chrisjsewell:
See also #362, https://github.com/executablebooks/meta/issues/306#issuecomment-827681278

chrisjsewell:
cc also @mgeier, in relation to: https://github.com/spatialaudio/nbsphinx/pull/560 (thanks also for the work on `mathjax3_config` 👍)

chrisjsewell:
> And BTW, please never use double dollars ($$) and don't recommend using them!

Why?

chrisjsewell:
> For me, the main reason is that there should be only one obvious way to do block equations, and that's math environments.

But then (a) you can't use them inline, and (b) a lot of people already use `$$` in Jupyter Notebook, not least their actual creator 😬: executablebooks/jupyter-book#1181

chrisjsewell:
(but I do understand your viewpoint)

[...]

PR #379, opened 2021-05-10
DOC: mention $$ instead of $

mgeier:
The current text suggests that MathJax recognizes single `$` by default, which is not true.


chrisjsewell:
Thnaks yeh I actually just remembered this whilst writing #378 lol

chrisjsewell:
As of #390, we now use the method of setting a top-level `mathjax_ignore` in myst documents and setting the MathJax config `"processClass": "tex2jax_process|mathjax_process|math"`.
Therefore, this is no longer relevant.

PR #390, opened 2021-06-11
⬆️ UPGRADE: sphinx v4 (& drop sphinx v2, change mathjax override)

chrisjsewell:
The main change is with mathjax. It's still a bit of a faff, but I think the approach is a little better: 
`tex2jax_process` and `mathjax_process` classes are added to the top-level sections of myst documents (this "fixes" having to override setting for RST documents), this tells mathjax not to render for any math under this section (i.e. the entire document) by default. The `math` class is then added to the process list, which overrides the default, and so tells mathjax to render elements (and child elements) with this class.
So basically it constricts mathjax to only rendering the elements we want.
This is also explained at; https://myst-parser--390.org.readthedocs.build/en/390/using/syntax.html#mathjax-and-math-parsing

As well as this, most of the sphinx test regression files have been split into sphinx3 (docutils<0.17) and sphinx4 (docutil 0.17) variants.

closes https://github.com/executablebooks/MyST-Parser/issues/378

### DIFF
--- a/myst_parser/mathjax.py
+++ b/myst_parser/mathjax.py
@@ -20,8 +20,12 @@ logger = logging.getLogger(__name__)
 
 
 def override_mathjax(app: Sphinx):
-    """Override aspects of the mathjax extension, but only if necessary."""
+    """Override aspects of the mathjax extension.
 
+    MyST-Parser parses dollar and latex math, via markdown-it plugins.
+    Therefore, we tell Mathjax to only render these HTML elements.
+    This is accompanied by setting the `ignoreClass` on the top-level section of each MyST document.
+    """
     if (
         "amsmath" in app.config["myst_enable_extensions"]
         and "mathjax" in app.registry.html_block_math_renderers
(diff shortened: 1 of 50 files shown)


mmcky:
🥳 -- thanks @chrisjsewell. Will do some testing. What's your timeframe for feedback (i.e. when do you want to merge this?)

chrisjsewell:
Well ASAP I guess for those waiting on it, can you look in the next day or 2?
Also @choldgraf?

myst_parser/mathjax.py
```diff
@@ -30,32 +34,34 @@ def override_mathjax(app: Sphinx):
...
+        app.config.mathjax3_config.setdefault("options", {})
+        if "processHtmlClass" in app.config.mathjax3_config["options"]:
+            logger.warning(
+                "`mathjax3_config['options']['processHtmlClass']` is being overridden "
+                f"by myst-parser to {mjax_classes}. "
+                "Set `myst_mathjax_classes = None` if this is undesirable."
+            )
+        app.config.mathjax3_config["options"]["processHtmlClass"] = mjax_classes
+    else:
+        # sphinx 3 + mathjax 2
```
choldgraf, line 55: Should we add "to do" items to deprecate this logic at some point? I dunno how many release cycles we'd want to support but just to make it easier for us to find future code to deprecate and clean up?

choldgraf (dismissed):
This looks great to me, just one minor under-the-hood thought but it's no biggie. Exciting to see support for v4!

Also I believe this will close https://github.com/executablebooks/MyST-Parser/issues/378 so added that to the top comment

chrisjsewell:
> Also I believe this will close #378 so added that to the top comment

FYI this was already in the linked issues. In case you didn't know adding in the initial comment is only useful when you first create the PR; to auto-add to the linked issues, after that it doesn't do anything

myst_parser/mathjax.py
chrisjsewell: Nah I thinks ok, since it specifically comments that it is for sphinx 3, and we would only remove if we drop sphinx 3 support

PR #404, opened 2021-07-08
📚  DOCS: Add big summary card

choldgraf:


damianavila:
Do you have a picture of how this one looks like on Twitter?

choldgraf:
I can't preview this PR directly on twitter because the ReadTheDocs preview blocks robot scraping, but here's what it looks like now:

Screen Shot 2021-07-08 at 4 36 15 PM

and here's what "large size" twitter cards look like (in this case it's a GitHub preview link):

Screen Shot 2021-07-08 at 4 38 02 PM

so for wide-style logos I think the large image summary is better

Issue #347, opened 2021-04-16
Split docutils only functionality into separate package


chrisjsewell:
extracted from https://github.com/executablebooks/MyST-Parser/issues/342:

> I was surprised to see docutils 0.17 add experimental Markdown support using Recommonmark

ah I completely missed the response on the email thread.

Ah well that's a different proposition: I literally have [myst_parser/docutils_renderer.py](https://github.com/executablebooks/MyST-Parser/blob/master/myst_parser/docutils_renderer.py) and [myst_parser/sphinx_renderer.py](https://github.com/executablebooks/MyST-Parser/blob/master/myst_parser/sphinx_renderer.py) as a sub-class, so that it is certainly possible to use myst-parser with "docutils only" functionality.

If I had known previously about such an intention to ship Markdown in docutils, I would certainly have considered splitting the "docutils only" aspects in to a separate "myst-docutils" package (and have that as a dependency here), which I think would probably be a better solution than just moving dependencies to extras.
Indeed, if there was an agreement in principle with the docutils guys to include something like myst-docutils in docutils I would look into this.

cpitclaudel:
> Ah well that's a different proposition: I literally have myst_parser/docutils_renderer.py and myst_parser/sphinx_renderer.py as a sub-class, so that it is certainly possible to use myst-parser with "docutils only" functionality.

I think this would be great.  The thing that's missing right now for convenient integration with docutils, I think (besides separating dependencies), is a version of `sphinx_parser` for docutils.   It would allow this kind of code to work:

```python
from docutils.core import publish_string
from myst_parser.sphinx_parser import MystParser
print(publish_string(source="{math}`e^{i\pi} = -1`", parser=MystParser(), writer_name='html5'))
```

(and with that integration into a variety of other docutils-based pipelines; my use case is integrating with [Alectryon](https://github.com/cpitclaudel/alectryon) for Coq proofs.)

As currently written, this code crashes:

```
Traceback (most recent call last):
  File "minimyst.py", line 3, in <module>
    print(publish_string(source="{math}`a^2 + b^2 = c^2`", parser=MystParser(), writer_name='html5'))
  File "/home/clement/.local/lib/python3.8/site-packages/docutils/core.py", line 407, in publish_string
    output, pub = publish_programmatically(
  File "/home/clement/.local/lib/python3.8/site-packages/docutils/core.py", line 665, in publish_programmatically
    output = pub.publish(enable_exit_status=enable_exit_status)
[...]

PR #406, opened 2021-07-19
[draft] Remove option mathjax_classes

cpitclaudel:
The global approach in #395 isn't really sustainable: it requires all-ways cooperation between all projects that want to customize MathJax.  Additionally, when processing a MyST document without Sphinx, the MathJax configuration changes are not performed (part of #347).  And, of course, this approach of overriding the MathJax object causes issues down the line for projects that need to customize MathJax (the setting in Sphinx isn't sufficient, see https://github.com/sphinx-doc/sphinx/issues/9450)

@chrisjsewell you wrote:

> Its not really ideal that either package has to override the (global) configuration

I think the following two approaches would not require overriding the global config:

1. Add a custom script instead of touching the mathjax3_config variable; something like this, essentially:

   ```js
   app.add_js_file(None, priority=0, body="""
      var MathJax = window.MathJax || MathJax;
      MathJax.options = MathJax.options || {};
      MathJax.options.processHtmlClass = (MathJax.options.processHtmlClass || "")
      + "|math";
   """)
   ```

2. Don't touch MathJax_config at all; instead, add an explicit `mathjax_process`  class on all math nodes, either by changing `docutils_renderer` (this PR) or by  adding a Docutils transform to processes all math nodes:

  ```python
  class ActivateMathJaxTransform(Transform):
      default_priority = 800

      @staticmethod
      def is_math(node):
          return isinstance(node, (math, math_block))

      def apply(self, **kwargs):
          for node in self.document.traverse(self.is_math):
              node.attributes.setdefault("classes", []).append("mathjax_process")
  ```

This PR isn't ready for merging; it's just to start a discussion.

### DIFF
--- a/myst_parser/docutils_renderer.py
+++ b/myst_parser/docutils_renderer.py
@@ -470,6 +470,13 @@ class DocutilsRenderer(RendererProtocol):
         self.add_line_and_source_path(node, token)
         self.current_node.append(node)
 
+    @property
+    def blocks_mathjax_processing(self):
+        return self.sphinx_env is None or (
+            "myst_update_mathjax" in self.sphinx_env.config
+            and self.sphinx_env.config.myst_update_mathjax
+        )
+
     def render_heading(self, token: SyntaxTreeNode) -> None:
 
         if self.md_env.get("match_titles", None) is False:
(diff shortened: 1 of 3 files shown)


chrisjsewell:
> Don't touch MathJax_config at all; instead, add an explicit mathjax_process class on all math nodes, either by changing docutils_renderer (this PR)

Yeh but this would break any other extension that adds math and likely has not added explicitly the `mathjax_process` class. Therefore, I won;t be excepting this in its current form. Unfortunately docutils nor sphinx have really provided any good mechanism to deal with this

cpitclaudel:
I don't understand — isn't it the reverse?  What MyST *currently* does breaks extensions that use math.

chrisjsewell:
No because the configuration sets Mathias to run on all elements with the `math` class, which is added for all `math` and `math_blocks`

cpitclaudel:
Unless, of course, some other extensions modifies the Mathjax config (which is the issue at hand); and unless the document is processed with Docutils instead of Sphinx, too.  That's why I was suggesting to add mathjax_process on relevant nodes.  But it's probably OK to leave this unaddressed.

PR #409, opened 2021-08-06
👌 IMPROVE: Fail gracefully on non-dict front-matter

chrisjsewell:


myst_parser/docutils_renderer.py
```diff
@@ -609,6 +609,7 @@ def render_front_matter(self, token: SyntaxTreeNode) -> None:
         if not isinstance(token.content, dict):
             try:
                 data = yaml.safe_load(token.content)
+                assert isinstance(data, dict), "not dict"
             except (yaml.parser.ParserError, yaml.scanner.ScannerError) as error:
```
chrisjsewell, line 613:
```suggestion
            except (AssertionError, yaml.parser.ParserError, yaml.scanner.ScannerError) as error:
```

cpitclaudel:
Is this patch actually safe? `assert` isn't guaranteed to run.

chrisjsewell:
Yeh that’s why it’s caught by the try/except

cpitclaudel:
No, I mean `assert` isn't guaranteed to *run* (I don't mean to succeed; I mean that Python can skip it entirely if you run MyST with `-O`).  Maybe I'm misunderstanding the point of that test.

chrisjsewell:
> run MyST with -O

can you explain further what you mean by this, and when this would be the case thanks?

cpitclaudel:
If you run Python with the `-O` flag, it skips all assertions.  This is why AFAIK it's usually not recommended to put stateful code in assertions or to depend on assertions for control flow.

chrisjsewell:
Thanks for the explanation BTW; I'd "heard" that you should not use assert, and usually try to follow that, but didn't actually know why lol.
Out of interest, is there some concrete use-case where one would want to use the `-O` flag?

cpitclaudel:
Not [too](https://stackoverflow.com/questions/1693088/what-is-the-use-of-pythons-basic-optimizations-mode-python-o) [many](https://stackoverflow.com/questions/9929456/keeping-assert-despite-the-o-optimize-flag) AFAIK ^^

chrisjsewell:
ok cheers

PR #413, opened 2021-08-12
📚 DOCS: Add reference to MySTyc in landing page

astrojuanlu:
Ref: https://twitter.com/juanluisback/status/1419626473787469824

mystyc

Encouraged by @pradyunsg and @chrisjsewell 🚀

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,7 +30,8 @@ MyST provides [roles](syntax/roles) and [directives](syntax/directives), allowin
 **[Compatible with Sphinx](sphinx/index.md)** 📄
 ^^^
 MyST is inspired by Sphinx, and comes with [its own Sphinx parser](sphinx/index.md).
-[Write your Sphinx docs in Markdown](sphinx:usage/quickstart), or convert existing [RST to Markdown][rst-to-myst]!
+[Write your Sphinx docs in Markdown](sphinx:usage/quickstart), or convert existing [RST to Markdown][rst-to-myst]
+from the CLI or [using an interactive web interface][mystyc]!
 
 ---
 **[Hackable with Python](api/index.md)** 🐍
@@ -61,8 +62,6 @@ In addition, here are a few pointers to help you get started.
 **[](syntax/syntax.md)**: discusses major MyST syntax components.
 
 **[The Sphinx guide](sphinx/index.md)**: how to use MyST with your Sphinx documentation.
-
-**[MySTyc](https://mystyc.herokuapp.com/)**: an online reST to MyST conversion tool.
 ---
 :header: bg-myst-two
 
@@ -136,3 +135,4 @@ The MyST markdown language and MyST parser are both supported by the open commun
 [markdown-it-myst]: https://github.com/executablebooks/markdown-it-myst
 [markdown-it]: https://markdown-it.github.io/
 [rst-to-myst]: https://rst-to-myst.readthedocs.io
+[mystyc]: https://mystyc.herokuapp.com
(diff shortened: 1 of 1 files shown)


astrojuanlu:
Rendered version: https://myst-parser--413.org.readthedocs.build/en/413/

Screenshot 2021-08-12 at 13-18-48 MyST - Markedly Structured Text

chrisjsewell:
thanks for the cool product @astrojuanlu!

I'd note that higher up the page there is: `, or convert existing [RST to Markdown][rst-to-myst]`, obviously mystrc uses rst-to-myst, so perhaps these two links should be combined, something like `RestructuredText conversion: rst-to-myst is a CLI tool which can also be used interactively with mystrc`

astrojuanlu:
Thanks @chrisjsewell ! Not sure if this is what you had in mind, WDYT?

Screenshot 2021-08-12 at 18-28-59 MyST - Markedly Structured Text

chrisjsewell:
> Not sure if this is what you had in mind, WDYT?

yeh I think possibly so thanks.

We could put it lower down, where you had it before, in the "Get started with MyST" panel,
but for me I feel it's quite a nice selling point to people that already have sphinx documentation in RST, so I want it to be more prominent, i.e. higher up the page.

what do you think @choldgraf?

choldgraf (approved):
This looks good to me - I think it's nice to have it in a prominent spot, though not super invasive.

My one concern is maintenance of the service over time. What are the odds that in 3 months somebody will click the link and the heroku app will be down? If folks don't think that is a significant possibility then I'm +1 on the change

astrojuanlu:
The service is running on my personal Heroku account on the free dyno ( $0.00/month ). Out of the [1000 free dyno hours granted for verified accounts](https://devcenter.heroku.com/articles/free-dyno-hours), MySTyc has spent 9.68 hrs this month. I have no intention of taking it down anytime soon.

However, there is a chance someone might abuse the app, or that someday it suffers an orange-site-calypse. If the app exhausts the free hours, my plan is to just scale it down temporarily until the beginning of the following month. If this is not acceptable to folks, here are some ideas:

- Replace the Python process with JavaScript code, so the computation happens on the client and doesn't need a backend. I don't think https://github.com/executablebooks/markdown-it-myst can do what rst-to-myst can, but I would love to be proven wrong here.
- Spend some more hours shielding the backend from trivial DoS attacks, for example setting a maximum number of requests by IP or similar. I have no idea what are the best practices about this.
- Make the costs public and share the load somehow, if they are ever non-zero (unlikely)
- ?

We could also give the app a domain name, so if it ever moves away from Heroku, the URL is unchanged. It costs money upfront, doesn't solve the underlying issue.

All of this sounds like not very appealing for a weekend hack, but I understand that you want to have some confidence on what you put on the front page. I'm open to suggestions here :)

choldgraf:
Hey @astrojuanlu - sorry for the slow response on that latest point. In my opinion, we shouldn't block this PR on battle-testing that heroku app. It's useful and cool, and even if it goes down at some point, at worst people will just open an issue about it, and we can remove the link in the future.

So I think the next step here is to just implement the location swap that @astrojuanlu mentions here: https://github.com/executablebooks/MyST-Parser/pull/413#issuecomment-897786207

and then give it a whirl :-)

astrojuanlu:
No worries @choldgraf ! The swap was already done in the last commit I think :D

chrisjsewell:
Yeh that's good for me cheers

choldgraf (approved):
Sounds great - thanks @astrojuanlu for the improvement!

Issue #420, opened 2021-08-22
Build single pages with docutils (instead of Sphinx)


cpitclaudel:
### Describe the enhancement you'd like

I would like a way to configure MyST from `docutils.conf` instead of `conf.py`.

### Does this solve a specific problem?

I compile most of my documents using plain docutils, not Sphinx.  This is because Sphinx isn't great for single-page documents: the setup process for a new document is very heavy, and it's strongly geared towards producing books, not articles or reports.

On the other hand, Docutils works great for short to medium-length single-file documents.

Being able to use Markdown instead of reST would be nice, but as far as I can tell there's no way to configure MyST from `docutils.conf` instead of Sphinx' `conf.py`, so there's no way to enable Markdown extensions or configure MyST in any way when using the plain-Docutils pipeline.

### What alternatives exist?

I'm not aware of a good workaround at the moment.

choldgraf:
I updated the title to be a bit more general, since I think what you really want is to be able to build with docutils directly, instead of going to Sphinx via docutils. Is that right? (if so, configuration with `docutils.conf` is likely only one step of the solution, rather than the whole solution).

Also - I'm curious why it wouldn't be acceptable to just use Sphinx but with a theme that was suited for single pages. What are the major downsides there for you?

cpitclaudel:
>  Is that right?

Yep, precisely.  But I think you're too pessimistic :) The code in this repo is 95% of the way there — docutils.conf is really the last missing step (I know that because in Alectryon you can already convert markdown to html using MyST + docutils, and the only part that's not working is docutils.conf (modulo the bugs like #418, which is easily fixed, and #406, which is quite painful).

> Also - I'm curious why it wouldn't be acceptable to just use Sphinx but with a theme that was suited for single pages. What are the major downsides there for you?

The overhead of conf.py: I use Sphinx heavily for larger projects, but for single pages it's not convenient at the moment: it doesn't have a good command line interface to mirror conf.py, so you need to write a new conf.py for each document.  And docutils `publish_programmatically` API is actually pretty nice, once you get used to it.

cpitclaudel:
Also, of course, there's the hope that we could have MyST as part of the standard docutils pipeline at some point, as was discussed in https://github.com/executablebooks/MyST-Parser/issues/347

chrisjsewell:
Yep, very quickly, I certainly think it is feasible

cpitclaudel:
[...]

PR #426, opened 2021-09-17
✨ NEW: Add Docutils MyST config and CLI

cpitclaudel:
This is a follow-up to my previous PR (GH-418),  part of an effort (GH-420) to make myst_parser integrate better with Docutils.
This exposes (almost) all MyST configuration settings in docutils.conf, making it possible to configure myst_parser for use without Sphinx.

I've added a simple entry point to make it easier to experiment (the second commit on this branch).  Here's a quick demo:

```console
# We have access to the standard docutils command line interface:
$ python3 -m myst_parser.docutils_ -h
...
Usage
=====
  docutils_.py [options] [<source> [<destination>]]

Generates (X)HTML documents from standalone MyST sources.  Reads from <source>
(default is stdin) and writes to <destination> (default is stdout).  See
<http://docutils.sf.net/docs/user/config.html> for the full reference.

Options
=======
General Docutils Options
------------------------
...

MyST options
------------
--myst-commonmark-only=MYST_COMMONMARK_ONLY
--myst-enable-extensions=MYST_ENABLE_EXTENSIONS
--myst-dmath-allow-labels=MYST_DMATH_ALLOW_LABELS
--myst-dmath-allow-space=MYST_DMATH_ALLOW_SPACE
--myst-dmath-allow-digits=MYST_DMATH_ALLOW_DIGITS
--myst-dmath-double-inline=MYST_DMATH_DOUBLE_INLINE
--myst-update-mathjax=MYST_UPDATE_MATHJAX
--myst-mathjax-classes=MYST_MATHJAX_CLASSES
--myst-disable-syntax=MYST_DISABLE_SYNTAX
--myst-url-schemes=MYST_URL_SCHEMES
--myst-heading-anchors=MYST_HEADING_ANCHORS
--myst-footnote-transition=MYST_FOOTNOTE_TRANSITION
--myst-sub-delimiters=MYST_SUB_DELIMITERS
--myst-words-per-minute=MYST_WORDS_PER_MINUTE
...

# We can compile markdown documents directly:
$ printf "# Title\n:::{note}\nIt works!\n:::" > test.md
$ python3 -m myst_parser.docutils_ --myst-enable-extensions=colon_fence test.md 
...
<body>
<div class="tex2jax_ignore mathjax_ignore document" id="title">
<h1 class="title">Title</h1>

<div class="admonition note">
<p class="first admonition-title">Note</p>
<p class="last">It works!</p>
</div>
</div>
</body>
</html>
```

Remaining issues:

- Math doesn't work, because of GH-406.  I think it would be reasonable to include the fix from that PR (add a transform that adds mathjax_process where relevant), but I'm happy to consider alternatives.
- There's no documentation for CLI options, because the documentation is on the web but not in the code.
- I need help with writing tests, as I don't have time to understand how the tests work at the moment :/

### DIFF
--- a/myst_parser/docutils_.py
+++ b/myst_parser/docutils_.py
@@ -40,7 +40,14 @@ def _create_validate_tuple(length: int) -> Callable[..., Tuple[str, ...]]:
     return _validate
 
 
-DOCUTILS_UNSET = object()
+class Unset:
+    """A sentinel class for unset settings."""
+
+    def __repr__(self):
+        return "UNSET"
+
+
+DOCUTILS_UNSET = Unset()
 """Sentinel for arguments not set through docutils.conf."""
 
 
--- a/myst_parser/docutils_.py
+++ b/myst_parser/docutils_.py
@@ -1,8 +1,9 @@
 """A module for compatibility with the docutils>=0.17 `include` directive, in RST documents::
 
-   .. include::
+   .. include:: path/to/file.md
       :parser: myst_parser.docutils_
 """
+from contextlib import suppress
 from typing import Any, Callable, Iterable, List, Optional, Tuple, Union
 
 from attr import Attribute
@@ -129,8 +130,9 @@ def create_myst_config(settings: frontend.Values):
         if attribute.name in DOCUTILS_EXCLUDED_ARGS:
             continue
         setting = f"myst_{attribute.name}"
-        val = getattr(settings, setting)
-        delattr(settings, setting)
+        val = getattr(settings, setting, DOCUTILS_UNSET)
+        with suppress(AttributeError):
+            delattr(settings, setting)
         if val is not DOCUTILS_UNSET:
             values[attribute.name] = val
     values["renderer"] = "docutils"
--- a/myst_parser/docutils_.py
+++ b/myst_parser/docutils_.py
@@ -3,45 +3,45 @@
    .. include::
       :parser: myst_parser.docutils_
 """
-from typing import Iterable, Tuple, Union
+from typing import Any, Callable, Iterable, Tuple, Union
 
 from attr import Attribute
 from docutils import frontend, nodes
+from docutils.core import default_description, publish_cmdline
 from docutils.parsers.rst import Parser as RstParser
 from markdown_it.token import Token
 
 from myst_parser.main import MdParserConfig, default_parser
 
 
-def validate_int(
+def _validate_int(
     setting, value, option_parser, config_parser=None, config_section=None
-):
+) -> int:
+    """Validate an integer setting."""
     return int(value)
 
 
-def validate_tuple(length: int):
-    def validate(
+def _create_validate_tuple(length: int) -> Callable[..., Tuple[str, ...]]:
+    """Create a validator for a tuple of length `length`."""
+
+    def _validate(
         setting, value, option_parser, config_parser=None, config_section=None
     ):
-        l = frontend.validate_comma_separated_list(
+        string_list = frontend.validate_comma_separated_list(
             setting, value, option_parser, config_parser, config_section
         )
-        if len(l) != length:
-            MSG = "Expecting {} items in {}, got {}: {}."
-            raise ValueError(MSG.format(length, setting, len(l)))
-        return tuple(l)
+        if len(string_list) != length:
+            raise ValueError(
+                f"Expecting {length} items in {setting}, got {len(string_list)}."
+            )
+        return tuple(string_list)
 
-    return validate
+    return _validate
 
 
 DOCUTILS_UNSET = object()
 """Sentinel for arguments not set through docutils.conf."""
 
-DOCUTILS_OPTPARSE_OVERRIDES = {
-    "url_schemes": {"validator": frontend.validate_comma_separated_list},
-    # url_schemes accepts an iterable or ``None``, but ``None`` is the same as ``()``.
-}
-"""Custom optparse configurations for docutils.conf entries."""
 
 DOCUTILS_EXCLUDED_ARGS = (
     # docutils.conf can't represent callables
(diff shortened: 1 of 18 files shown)


chrisjsewell:
cheers, I’ll give a proper look soon

> There's no documentation for CLI options, because the documentation is on the web but not in the code.

It would be very possible to self document the `MDConfig` attr class via metadata. I’ve already done it in other projects: https://github.com/chrisjsewell/ase-notebook/blob/eb12723f0aed4e774484f0e8bce10b3785c0a2e5/ase_notebook/configuration.py#L114

> I need help with writing tests, as I don't have time to understand how the tests work at the moment :/

is it ok if I push directly to your branch?

cpitclaudel:
> is it ok if I push directly to your branch?

Please do, thanks a lot!

cpitclaudel:
> It would be very possible to self document the MDConfig attr class via metadata.

Yep, that would be ideal.  Ideally that would also be used to generate the relevant section of the docs, to avoid having two sources of truth.  I just didn't find the time to do it ^^

chrisjsewell:
Ok @cpitclaudel, how's this!

Now you have separate CLI commands for writers: `myst-docutils-html` ` myst-docutils-html5` `myst-docutils-latex` `myst-docutils-pseudoxml` `myst-docutils-xml`

and full doc-strings:

```
$ myst-docutils-xml --help
Usage
=====
  myst-docutils-xml [options] [<source> [<destination>]]

Generates Docutils-native XML from standalone MyST sources. Reads from
<source> (default is stdin) and writes to <destination> (default is stdout).
See <http://docutils.sf.net/docs/user/config.html> for the full reference.

Options
=======
...
MyST options
------------
--myst-commonmark-only=MYST_COMMONMARK_ONLY
                        Use strict CommonMark parser (type: bool, default:
                        False)
--myst-enable-extensions=MYST_ENABLE_EXTENSIONS
                        Enable extensions (type: comma-delimited, default:
                        'dollarmath')
--myst-dmath-allow-labels=MYST_DMATH_ALLOW_LABELS
                        Parse `$$...$$ (label)` (type: bool, default: True)
--myst-dmath-allow-space=MYST_DMATH_ALLOW_SPACE
                        dollarmath: allow initial/final spaces in `$ ... $`
                        (type: bool, default: True)
--myst-dmath-allow-digits=MYST_DMATH_ALLOW_DIGITS
                        dollarmath: allow initial/final digits `1$ ...$2`
                        (type: bool, default: True)
--myst-dmath-double-inline=MYST_DMATH_DOUBLE_INLINE
                        dollarmath: parse inline `$$ ... $$` (type: bool,
                        default: False)
--myst-disable-syntax=MYST_DISABLE_SYNTAX
                        Disable syntax elements (type: comma-delimited,
                        default: '')
--myst-url-schemes=MYST_URL_SCHEMES
                        URL schemes to allow in links (type: comma-delimited,
                        default: 'http,https,mailto,ftp')
--myst-footnote-transition=MYST_FOOTNOTE_TRANSITION
                        Place a transition before any footnotes (type: bool,
                        default: True)
--myst-words-per-minute=MYST_WORDS_PER_MINUTE
                        For reading speed calculations (type: int, default:
                        200)
```

I removed the addition of the section level mathjax classes, when using docutils (since this is only useful with sphinx.ext.mathjax)

I also removed the `heading_anchors` option, since heading anchor slugs are currently only handled by the sphinx renderer.
It might be nice to eventually allow this functionality for docutils

cpitclaudel:
😍

cpitclaudel:
I forget if the `heading_anchors` option is related to this recently implemented feature in docutils: https://sourceforge.net/p/docutils/feature-requests/28/

cpitclaudel:
On the topic of script names, since rst uses `rst2XXX.py`, maybe we should have `myst2xxx.py` for consistency?
Although see https://sourceforge.net/p/docutils/feature-requests/82/:

> Docutils 0.17 comes with tools/docutils-cli.py:
> New generic command line front end that allows the free selection of
> reader, parser, and writer components.

chrisjsewell:
> maybe we should have myst2xxx.py for consistency

I was hesitant to have this, since it does not make it clear that it will only parse using docutils and not sphinx, e.g. you can't properly parse any files that use sphinx/sphinx-extension roles/directives

chrisjsewell:
> Although see https://sourceforge.net/p/docutils/feature-requests/82/

I very much agree with this issue though; it's a shame there is currently no easy "middle ground" between the single document docutils API/CLI, lacking any support for sphinx features, and the sphinx-build CLI, requiring use of the full, complex, project build system

myst_parser/docutils_.py
```diff
@@ -1,36 +1,157 @@
...
+def create_myst_config(settings: frontend.Values):
+    """Create a ``MdParserConfig`` from the given settings."""
+    values = {}
+    for attribute in MdParserConfig.get_fields():
+        if attribute.name in DOCUTILS_EXCLUDED_ARGS:
+            continue
+        setting = f"myst_{attribute.name}"
+        val = getattr(settings, setting, DOCUTILS_UNSET)
+        with suppress(AttributeError):
+            delattr(settings, setting)
```
chrisjsewell, line 135: @cpitclaudel why do you need to delete the setting here?

chrisjsewell: (note here I made the getting/deleting optional, since if you include a MyST file from an RST file, the MyST settings will not have been loaded)

chrisjsewell:
> I forget if the heading_anchors option is related to this recently implemented feature in docutils: sourceforge.net/p/docutils/feature-requests/28

If you want to open this as a future improvement issue

myst_parser/docutils_.py
cpitclaudel: I forget the details :/ I *think* it was to avoid namespace pollution? I'm guessing I was worried about having the settings both directly in the document under `myst_*` and in a config object, so I was planning to remove all the `myst_*` settings from `document.settings` and instead have a single object `document.settings.myst_config`.

Of course, this explanation would make more sense if I had then actually stored the config object as `document.settings.myst_config`, but the code doesn't seem to do that, so either I forgot or my guess above is wrong ^^

chrisjsewell:
Documentation added 😄 https://myst-parser--426.org.readthedocs.build/en/426/docutils.html

myst_parser/docutils_.py
chrisjsewell: I think I'll remove it for now then, as it doesn't appear to break anything

chrisjsewell:
Ok I think this is good to go! @cpitclaudel do you want to review your own PR lol, and let me know if you are happy for it to be merged

chrisjsewell (approved):

chrisjsewell:
@cpitclaudel let me know if you happy for this to be merged cheers, eager to get out version 0.16.0 😄 (#426 -> #456 -> #457 -> 🚀)

chrisjsewell:
Ok times up 😛 but I'm sure its all fine

cpitclaudel:
> Ok times up but I'm sure its all fine

Yep!  Thanks a lot for moving ahead, I'm terrible at email and super busy these days.  Very excited to see things moving in this direction :+1:

PR #456, opened 2021-12-06
🔧 MAINTAIN: Add publishing job for `myst-docutils`

### LINKED ISSUE #347: Split docutils only functionality into separate package

chrisjsewell:
This PR adds a workflow for releasing an additional package `myst-docutils`, which does not include install requirements on docutils or sphinx, to close #347 and allow docutils to utilise it in their package with no cyclic dependencies.
It does this by simply modifying the `setup.cfg` and `README.md` before the build.
The build is also tested against a range of docutils versions (e.g. to ensure it does not fail due to missing sphinx install)

### DIFF
--- a/.github/workflows/docutils_setup.py
+++ b/.github/workflows/docutils_setup.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+"""Script to convert package setup to myst-docutils."""
+import configparser
+import io
+import sys
+
+
+def modify_setup_cfg(content: str) -> str:
+    """Modify setup.cfg."""
+    cfg = configparser.ConfigParser()
+    cfg.read_string(content)
+    # change name of package
+    cfg.set("metadata", "name", "myst-docutils")
+    # move dependency on docutils and sphinx to extra
+    install_requires = []
+    sphinx_extra = [""]
+    for line in cfg.get("options", "install_requires").splitlines():
+        if line.startswith("docutils"):
+            sphinx_extra.append(line)
+        elif line.startswith("sphinx"):
+            sphinx_extra.append(line)
+        else:
+            install_requires.append(line)
+    cfg.set("options", "install_requires", "\n".join(install_requires))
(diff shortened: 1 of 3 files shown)


chrisjsewell:
will adapt this slightly once #426 is merged

.github/workflows/docutils_setup.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def modify_readme(content: str) -> str:
+    """Modify README.md."""
+    content = content.replace(
+        "# MyST-Parser",
+        "# MyST-Parser\n\nNote: myst-docutils is identical to myst-parser, "
+        "but without installation requirements on sphinx",
+    )
+    content = content.replace("myst-parser", "myst-docutils")
```
astrojuanlu, line 39: Because of this replacement, the Readme now says "myst-docutils is identical to myst-docutils" 😄

chrisjsewell: Doh!

Issue #435, opened 2021-10-01
file:///// links skipped in HTML output of Sphinx


Mershl:
### Describe the problem

The behaviour of an inline link to an external resource (access through file://///) does not work as expected for an .md input to MyST when generating in HTML through Sphinx. (e.g. for referencing files on network shares)

Given an inline link like:
```
[Click me](file://///foo.com/dataXYZ/documentation.pdf)
```

Results in the inline link being output unprocessed:
![grafik](https://user-images.githubusercontent.com/520251/135624474-63f623e4-1136-4c4b-a6b6-7b74931253fb.png)

Creating the link manually in the built html using:
```
<a class="reference external" href="file://///foo.com/dataXYZ/documentation.pdf">Click me</a>
```

Results in the expected link and works as expected in our tested browsers.

I guess that file:///// links are disabled/ignored by MyST per default as the protocol is disabled by most modern web browsers. In companies this legacy feature often times still used and enabled for the used browser. Is it possible to enable MyST to generate the expected link for this input?

### Steps to reproduce

Using Sphinx and MyST.

1. Create an .md containing:
`[Click me](file://///foo.com/dataXYZ/documentation.pdf)`
2. Add the .md to your index.rst
3. Call ./make html to start the HTML generation of Sphinx
4. Open generated .html. The file will contain the link unprocessed, looking like this:
![grafik](https://user-images.githubusercontent.com/520251/135624474-63f623e4-1136-4c4b-a6b6-7b74931253fb.png)

### The version of Python you're using

3.9.7

### Your operating system

Windows 10

### Versions of your packages

Sphinx==4.0.2
myst-parser==0.15.1
sphinx-rtd-theme==0.5.2

chrisjsewell:
[...]

Issue #411, opened 2021-08-10
“'myst' reference target not found: b.md#anchor” when building in parallel


andersk:
### Describe the problem

I configured `myst_heading_anchors = 2` as [documented](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-header-anchors), but anchor links are giving me this warning *if I enable parallelism* (`sphinx-build -W -j N` for any `N` other than 1):

```
Warning, treated as error:
/tmp/myst-anchor-test/docs/dir/a.md:2:'myst' reference target not found: b.md#anchor
```

When building without parallelism, there is no warning and the anchor works fine.

### Link to your repository or website

https://github.com/andersk/myst-anchor-test

### Steps to reproduce

```console
$ pip install Sphinx myst-parser
…
Successfully installed Jinja2-3.0.1 MarkupSafe-2.0.1 Pygments-2.9.0 Sphinx-4.1.2 alabaster-0.7.12 attrs-21.2.0 babel-2.9.1 certifi-2021.5.30 charset-normalizer-2.0.4 docutils-0.17.1 idna-3.2 imagesize-1.2.0 markdown-it-py-1.1.0 mdit-py-plugins-0.2.8 myst-parser-0.15.1 packaging-21.0 pyparsing-2.4.7 pytz-2021.1 pyyaml-5.4.1 requests-2.26.0 snowballstemmer-2.1.0 sphinxcontrib-applehelp-1.0.2 sphinxcontrib-devhelp-1.0.2 sphinxcontrib-htmlhelp-2.0.0 sphinxcontrib-jsmath-1.0.1 sphinxcontrib-qthelp-1.0.3 sphinxcontrib-serializinghtml-1.1.5 urllib3-1.26.6
$ git clone https://github.com/andersk/myst-anchor-test.git
$ cd myst-anchor-test
$ sphinx-build -W -j2 docs build
Running Sphinx v4.1.2
…
Warning, treated as error:
/tmp/myst-anchor-test/docs/dir/a.md:2:'myst' reference target not found: b.md#anchor
$ sphinx-build -W docs build
Running Sphinx v4.1.2
build succeeded.

The HTML pages are in build.
```

If you want to reproduce again, `rm -rf build`.

### The version of Python you're using

3.9.6

### Your operating system

NixOS 21.11

### Versions of your packages

```console
$ pip freeze
alabaster==0.7.12
attrs==21.2.0
Babel==2.9.1
certifi==2021.5.30
[...]

Issue #402, opened 2021-06-26
Can


jaymegordo:
```
/Users/Jayme/OneDrive/MDS/591/Orbis-Investments-Capstone/CONTRIBUTING.md:4: WARNING: 'myst' reference target not found: README.md
```

**To Reproduce**

**Expected behavior**

A link to your documentation repository.

If relevant, a minimal example of the input text should be supplied,
together with a screen-shot of the output document and/or command-line output, e.g.

```markdown
some text...
```

```console
$ jupyter-book build mybook
ERROR ...
```

**Environment**

- Python Version [e.g. 3.7.1]:
- Package versions or output of `jupyter-book --version`:
- Operating System:

**Additional context**

chrisjsewell:
heys @jaymegordo could you provide a minimal working example to reproduce this warning, i.e. it seems like you have something like this

```
docs/source/conf.py
    extensions = ["myst_parser"]
docs/source/index.md
    content?
CONTRIBUTING.md
    # Heading

    [README](README.md)
README.md
    # Heading
```

vineetsharma883:
I have following folder structure:-
```
automation_p/
    |_____worfklow.md
    |_____Conditional.sh
docs/
    |____source
    |     |_____conf.py
    |     |_____index.rst
    |_____build
```

Following is problematic line in `workflow.md` - 
```
...
...
to check the condition ref [Conditional.sh](./Conditional.sh)
...
...
```

On running `make html`, I am getting following error - `WARNING: 'myst' reference target not found: ./Conditional.sh`

choldgraf:
I believe that the issue here is that `.sh` files are not supported by Sphinx, and so it cannot find it. If you want to link to a raw file, use ``{download}`./Conditional.sh <Conditional.sh>` ``

arwedus:
I guess that the need of the original author is to suppress these warnings, not to fix the issue itself.
I have the same request.

I include some files in the Sphinx documentation which are targeted at people who navigate the source code in an IDE or in a github or Bitbucket GUI. I include links to certain files which should not be copied to the documentation output, e.g. a header file or source file, to support easy navigation in the code.

The warnings we get are along the lines of:

```
/home/user/workspace/package/README.md:13: WARNING: 'myst' reference target not found: ./CMakeLists.txt
/home/user/workspace/package/README.md:13: WARNING: 'myst' reference target not found: ./src/main.cpp
```

Wanted behavior: Just remove the link and output the text (if available) or the link.

[...]

Issue #193, opened 2020-08-03
Configure external link recognition (and handling)


chrisjsewell:
Currently, if a link (`[text](link)`) does not match a URL scheme (e.g. 'http://...') then it is treated as an internal cross-reference (to a reference target or sphinx document, etc):
https://github.com/executablebooks/MyST-Parser/blob/3d5ae4f94c9d39435d76861b86dc5171ee23c9df/myst_parser/docutils_renderer.py#L412-L415

In some use cases a configuration option could be useful, such that links with certain extensions (or regexes) are converted to external links, rather than attempting to resolve them as internal links

_Originally posted by @chrisjsewell in https://github.com/executablebooks/jupyter-book/issues/823#issuecomment-668135898_

jpmckinney:
Same as #215?

chrisjsewell:
yep I guess so, and more or less #361

chrisjsewell:
although with #215 it is more than just leaving it as an external link, but turning it into a download link.
the focus though is: more configurable handling of the value of the `(link)`

Griatch:
`recommonmark` provided a `url_resolver` feature, where one could provide a custom handler to pre-process whatever is parsed as the url. This made it very easy to make custom urls be parsed into whatever end result. Including stuff like `[mylink](github:issues)` being parsed into an external link to the github issue page, or `[apilink](api:/foo/bar)` becoming the correct full link to the api documentation page.

chrisjsewell:
> recommonmark provided a url_resolver feature, where one could provide a custom handler to pre-process whatever is parsed as the url.

errr, it may be possible to do that here.
But, I maybe fell you are stepping away from what a Markdown link actually is, and there are other ways to achieve this.
For example, you could use a [substitution](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2) for the link to github issues, like:

```[ython
myst_substitutions = {
  "gh_issues": "[GitHub Issues](https://github.com/executablebooks/MyST-Parser/issues)"
}
```

Or you could create a new sphinx role for special links: ``` {api}`foo/bar` ```

Griatch:
@chrisjsewell 
While one can achieve the outcomes like you suggest, it requires quite a lot of changes to (in my case) vast amounts of already written Markdown documentation using `recommonmark`. Even though I suppose I could write scripts to try to convert, it's not really an ideal solution: 

[...]

PR #467, opened 2021-12-17
♻️ REFACTOR: Parsing logic of Markdown links

chrisjsewell:
This PR aims to clarify and improve the parsing of Markdown links, i.e. `[Text](link)`

Related issues: #193, #402, #411, #435

## Current logic

1. For both docutils and sphinx, if the link starts with a scheme prefix (e.g. `http:`) and that scheme is in `myst_url_schemes` or `myst_url_schemes=None`, then create an external link, e.g. `<a href="http://example.com">Text</a>
2. For both docutils and sphinx, if the link contains a `#` (i.e. a fragment), then also create an external link (unless the `heading_anchors` extension is active)
3. If docutils, and neither of the above, then nothing is rendered, and a warning is raised for a missing reference
4. If sphinx, and neither of the above, then create a sphinx node `addnodes.pending_xref` of type "myst"

After (4), the "myst" `pending_xref` are then intercepted in the `MystReferenceResolver` post-transform (after all documents have been parsed), and the reference resolved similar to the sphinx [`any` role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-any):
They check the reference against all reference types, taking the first hit as the target and emitting a warning, if none or multiple targets are found.
Different to the standard `any` role sphinx `ReferenceResolver`, the myst one allows for (a) matching document paths with their extensions, e.g. `[Text](path/to/other.md)`, and (b) parsing nested syntax as part of the text, e.g. `[Text **bold**](link)`.

there is also a little more complication, when using the `heading_anchors` extension, to match generated header targets, e.g. `[Text](document.md#header-1)`

## Shortcomings

(2) I believe originally came from recommonmark, but I think it is probably too permissive and unexpected.

(3) Is definitely not ideal, for docutils only use

(4) What is currently not handled, is links to "non-sphinx" files, e.g. non RST or Markdown files like `[Text](path/to/file.txt)`. In sphinx we would want to use the `download` role for this, to ensure that the referenced files are copied into the build directory and the end `href` points to that.

## Current Improvements

This PR is in draft, but current improvements are:

- Add a configuration `myst_all_links_external=True`, which simply directs for all links to be rendered as external links.
  I think this can be helpful for especially docutils use
- Remove step (2)
- For step (3) create a docutils internal link
- In step (4), if the link matches a "non-sphinx" file path, then create a `download_reference` node.

Perhaps there can be some more configuration,
but one thing I want to balance is (a) having sensible defaults, (b) allowing for some configuration, but (c) not having too much allowed configuration, which would be confusing to basic users and also then need to be replicated in any future MyST parser implementations.

## Questions

- Is it a problem to remove (2); no longer assuming links with `#` are external links?

I'll expand on this a bit more at a later data, but welcome comments from interested parties: @choldgraf, @mmcky, @rowanc1, @cpitclaudel, etc


choldgraf:
This sounds pretty reasonable to me. This sounds like a nice improvement. A few thoughts:

- Agree that `#2` feels like we are trying to make too many assumptions on the user's behalf.
- Agree that when people link directly to a file, the right behavior to assume is the `{download}` role
- Presumably `myst_all_links_external=True` defaults to `False`, at least for when we're in a Sphinx context?

chrisjsewell:
> Presumably myst_all_links_external=True defaults to False, at least for when we're in a Sphinx context?

yes, `myst_all_links_external` default is `True`

chrisjsewell:
See full documentation: https://myst-parser--467.org.readthedocs.build/en/467/syntax/syntax.html#markdown-links-and-referencing

choldgraf:
👍

mmcky:
thanks @chrisjsewell -- I think this is a nice improvement. 

I particularly like that links to non-source documents are migrated to download links. I think that is nice behaviour. 👍

unparalleled-js:
I used to be able to link to local HTML (post-build) files but it seems that is no longer the case?

PR #492, opened 2022-01-06
✨ NEW: Add `myst_title_to_header` configuration

chrisjsewell:
This is the "simplest" implementation (3 extra lines!).

Basically, it treats this:

```markdown
---
title: "A **title**"
---
```

like this (including parsing nested syntax)

```markdown
# A **title**
```

I'm not sure if this would be the correct/expected implementation though

for example, currently

```markdown
---
title: "A **title**"
---

# Another title
```

would be treated as

```markdown
# A **title**

# Another title
```

but should it instead be treated as

```markdown
# A **title**

## Another title
```

(also, on a technical note, the parsing should be a bit smarter, in only allowing inline syntax,
there should probably also be a config to turn it off/on)

### DIFF
--- a/docs/sphinx/reference.md
+++ b/docs/sphinx/reference.md
@@ -32,6 +32,9 @@ To do so, use the keywords beginning `myst_`.
 * - `myst_linkify_fuzzy_links`
   - `True`
   - If `False`, only links that contain a scheme (such as `http`) will be recognised as external links.
+* - `myst_title_to_header`
+  - `False`
+  - If `True`, the `title` key of a document front-matter is converted to a header at the top of the document.
 * - `myst_heading_anchors`
   - `None`
   - Enable auto-generated heading anchors, up to a maximum level, [see here](syntax/header-anchors) for details.
--- a/docs/syntax/syntax.md
+++ b/docs/syntax/syntax.md
@@ -398,8 +398,26 @@ This is a YAML block at the start of the document, as used for example in
 
 :::{seealso}
 Top-matter is also used for the [substitution syntax extension](syntax/substitutions),
-and can be used to store information for blog posting (see [ablog's myst-parser support](https://ablog.readthedocs.io/manual/markdown/)).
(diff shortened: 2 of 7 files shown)


choldgraf:
Agree with the implementation + your assessment of where it might be confusing. I think it'd be great if specifying `title:` also bumped all other headers on the page down one level. However if the implementation is complex, then just this PR as it already stands would be an improvement I think

chrisjsewell:
I think it would be good to check what the behaviour of e.g. pandoc is on this, then just copy that. If even a little more complex, it should be possible either way

chrisjsewell:
> behaviour of pandoc

See https://pandoc.org/MANUAL.html#metadata-blocks

>  when the `--standalone` (`-s`) option is chosen .. The title in the body appears as an H1 element

`test.md`
```markdown
---
title: A **title**
---

# Header

```

```console
$ pandoc -s test.md
...
<body>
<header id="title-block-header">
<h1 class="title">A <strong>title</strong></h1>
</header>
<h1 id="header">Header</h1>
</body>
```

i.e. the `title` is parsed as Markdown, and does not affect subsequent headings

chrisjsewell:
Ok good to go, if you agree @choldgraf, @rowanc1

rowanc1:
Looks good to me! Good to notice the option for this, and updating the docs!

chrisjsewell:
> Looks good to me! Good to notice the option for this, and updating the docs!

Cheers!

Note, with the recent updates I've made, it would actually be quite easy to set a "header level offset", i.e. to turn every `#` into `##` etc.
But as discussed, this is not the behaviour of pandoc (and thus also quarto) plus, as noted by @choldgraf (https://github.com/executablebooks/meta/discussions/615#discussioncomment-1921714)

> Markdown tends to map `#` to `<h1>` and `##` to `<h2>`

which I agree is easy for users to understand.

So I think the current behaviour is correct and no offset is needed

PR #501, opened 2022-01-09
✨ NEW: Add `myst_gfm_only` configuration

chrisjsewell:
For parsing of only [GitHub-flavored Markdown](https://github.github.com/gfm/)


chrisjsewell:
This is also intended for downstream use, in rendering Markdown code ouput in myst-nb/jupyter-book

Issue #399, opened 2021-06-23
Optional, but on by default syntax


hukkin:
Why does MyST have optional syntax that is on by default. At least dollarmath is like this, and it feels counter intuitive to me. I'd expect anything that is on by default to be part of "core" MyST syntax.

If nothing else, I think this sentence should be fixed in the docs:
> The following syntaxes are optional (disabled by default)

chrisjsewell:
See https://github.com/executablebooks/mdformat-myst/pull/9#issuecomment-867055932;
It’s only dollarmath and that’s mainly for legacy reasons and a deprecation pathway is a bit of a pain

choldgraf:
Yeah - to @chrisjsewell's point this is one reason that we moved dollarmath to the "optional" page in the latest docs refactor (#385 I believe). IMO it is less confusing to have an "optional syntax" that is "on by default" rather than to document an "optional syntax" in the "core syntax" page.

Maybe we should just add a quick note about this here: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#math-shortcuts

Something like

```
:::{note}
Dollar math is **on by default** even though it is an optional syntax.
This is because of its common use across many markdown formatters, and for legacy purposes.
We may revisit whether to consider this as part of "core" syntax in the future.
:::
```

?

PR #385, opened 2021-05-26
📚 DOCS: Refactoring and updating documentation

choldgraf:
This PR overhauls much of the MyST documentation, with the following goals:

- Separate out the "Sphinx extension", "MyST Specification", and "Python API" parts of this package more cleanly
- Take inspiration from the [diataxis framework](https://diataxis.fr/) for structuring our content
- Provide a bit more explanation etc around our various examples
- Add a "getting started tutorial" that is more like a step-by-step guide to using MyST in Sphinx.

My hope is that this will overall make the documentation easier to discover/parse, understand, and potentially extend in the future as we wish to add more information.

Note that there is still a lot of improvement we could make to the documentation, but I wanted to leave it here as a mostly "structural change" and we can discuss ways to improve it further in the future.

closes #383 

ping to @damianavila as well, perhaps he has thoughts on whether this structure would make it easier for newcomers to understand MyST and the various ways one can use it.

Probably the easiest way to review this is to just look at the built docs preview: https://myst-parser--385.org.readthedocs.build/en/385/

PR #505, opened 2022-01-14
‼️ BREAKING: Remove dollarmath from default `myst_enable_extensions`

chrisjsewell:
Also, the `tex2jax_ignore mathjax_ignore` classes are only added to the document if the `dollarmath` extension is enabled.
Similarly, the mathjax configuration is only overriden if the `dollarmath` extension is enabled.

### References

- Discussed in #399
- Some steps in this direction made in #385


choldgraf (commented):
This seems reasonable to me - if I recall, the major discussions were around how supporting dollar math creates a bunch of headaches for parsers and occasionally unexpected behavior in the results?

A couple of quick questions:

- I guess this means that we are changing the "core MyST spec" since we are removing a syntax that was on by default. That's probably fine now since MyST hasn't hit an official "stable" spec yet, but I just wanted to confirm that you all agree that this is what we are doing here. (in the future, this is the kind of thing that an enhancement proposal process might be helpful for)
- Does this mean that the recommended way for `MyST Markdown` to parse math is via a `{math}` directive? If so, perhaps that directive should be part of the "core" MyST spec, or alternatively we could decide that MyST doesn't treat mathematics in a special way at all in its core spec (not in this PR, just a note for the future)?
- Did we have any of these discussions in a public space that we could cross reference? I can take a look when I have a moment next week but just wanted to mention it in case others know of issues/discussions/etc off-hand

mmcky:
@chrisjsewell @choldgraf -- sorry for my slow response, was on the road yesterday for 13 hours so offline. This is a really useful syntax for a lot of authors. I don't mind if it get's removed from `myst-parser` for `stability` reasons. However (as you say) I would recommend we leave it enabled it by default for `jupyter-book`. 👍

chrisjsewell:
> Did we have any of these discussions in a public space that we could cross reference?

Yes this also was already discussed in #399, and you essentially already made part of this change in #385

> I guess this means that we are changing the "core MyST spec" 

dollarmath had already been moved to the optional syntax docs page in #385, so this is essentially already the case.

> Does this mean that the recommended way for MyST Markdown to parse math is via a `{math}` directive?

This, in part, depends on what level of compatibility you want with "native" Jupyter Notebooks.
`{math}` obviously does not "degrade" as gracefully as using `$` math, in this context.

It is kind of annoying that both sphinx and jupyter load mathjax by default, whose default configuration then "aggressively" converts all `$` in the HTML to math.
This kind of gives the false impression that dollarmath is being directly supported by MyST, even if not enabled, and obviously the mathjax approach does not support non-HTML output formats
(up-front `$` identification in the Markdown parser, is most definitely the better solution, which is what the `dollarmath` extension does)

> I would recommend we leave it enabled it by default for jupyter-book

It does not affect jupyter-book, which "hard-codes" the default MyST extensions enabled

choldgraf (approved):
Looks good to me 👍 

A few quick responses:

> Yes this also was already discussed in #399, and you essentially already made part of this change in #385

Thanks for digging up those links. I've updated the top comment with refs to those issues - just  wanted to follow best practices to link back to conversations about stuff.

> up-front $ identification in the Markdown parser, is most definitely the better solution, which is what the dollarmath extension does

If Jupyter moved away from marked.js and towards markdown-it, I wonder if this could be handled more gracefully by using a dollarmath extension in markdown-it, that would then render the right syntax for MathJax to process 🤔

chrisjsewell:
> Thanks for digging up those links. I've updated the top comment with refs to those issues - just wanted to follow best practices to link back to conversations about stuff.

cheers

> If Jupyter moved away from marked.js and towards markdown-it, I wonder if this could be handled more gracefully by using a dollarmath extension in markdown-it, that would then render the right syntax for MathJax to process 🤔

Yep exactly; the "problem" at present, it that MathJax is acting as both a parser and a renderer,
Ideally, all the parsing (i.e. syntax tokenization) should be performed by the Markdown parser, then it just utliises mathjax/katex to turn the LaTeX into HTML. 

For example, this is what https://github.com/executablebooks/markdown-it-dollarmath does

On the "python side", even more ideal, would be that you had a python implementation of mathjax/katex. Then the static HTML you generate would already have the math in HTML, and you wouldn't need to be running "dynamic" JavaScript.
This is essentially what happens with code blocks, that get converted to HTML by https://pygments.org/ during the sphinx/docutils build

PR #520, opened 2022-02-15
🐛 FIX: `parse_directive_text` when body followed by options

chrisjsewell:
This was uncovered in https://github.com/executablebooks/markdown-it-docutils/pull/28

The change does not break any existing tests, but I should add a test that would break without this change


rowanc1:
So cool to see the JS implementation pushing on the python side. Love it. Thanks for making these changes here!

chrisjsewell:
superseded by #580

Issue #522, opened 2022-02-19
Test failure on 32bit platforms: test_sphinx_directives[35-highlight (`sphinx.directives.code.Highlight`):]


stefanor:
### Describe the bug

**context**
As discovered in Debian unstable: https://bugs.debian.org/1005691

When running the test suite on a 32-bit platform, `test_sphinx_directives[35-highlight (sphinx.directives.code.Highlight):]` fails.

```console
=================================== FAILURES ===================================
__ test_sphinx_directives[35-highlight (`sphinx.directives.code.Highlight`):] __

line = 35, title = 'highlight (`sphinx.directives.code.Highlight`):'
input = '```{highlight} something\n```\n'
expected = '<document source="notset">\n    <highlightlang force="False" lang="something" linenothreshold="9223372036854775807">\n'

    @pytest.mark.parametrize(
        "line,title,input,expected",
        read_fixture_file(FIXTURE_PATH.joinpath("sphinx_directives.md")),
        ids=[
            f"{i[0]}-{i[1]}"
            for i in read_fixture_file(FIXTURE_PATH / "sphinx_directives.md")
        ],
    )
    def test_sphinx_directives(line, title, input, expected):
        # TODO fix skipped directives
        # TODO test domain directives
        if title.startswith("SKIP"):
            pytest.skip(title)
        elif title.startswith("SPHINX3") and sphinx.version_info[0] < 3:
            pytest.skip(title)
        elif title.startswith("SPHINX4") and sphinx.version_info[0] < 4:
            pytest.skip(title)
        document = to_docutils(input, in_sphinx_env=True)
        _actual, _expected = [
            "\n".join([ll.rstrip() for ll in text.splitlines()])
            for text in (document.pformat(), expected)
        ]
        try:
>           assert _actual == _expected
E           assert '<document so..."2147483647">' == '<document so...36854775807">'
E             Skipping 84 identical leading characters in diff, use -v to show
E             - hreshold="9223372036854775807">
E             + hreshold="2147483647">

tests/test_renderers/test_fixtures.py:121: AssertionError
----------------------------- Captured stdout call -----------------------------
<document source="notset">
    <highlightlang force="False" lang="something" linenothreshold="2147483647">

=========================== short test summary info ============================
FAILED tests/test_renderers/test_fixtures.py::test_sphinx_directives[35-highlight (`sphinx.directives.code.Highlight`):]
```

### Reproduce the bug

[...]

PR #523, opened 2022-02-19
🧪 TESTS: Fix regression output on 32-bit systems to match 64-bit systems

stefanor:
Fixes `test_sphinx_directives[35-highlight (sphinx.directives.code.Highlight):]` which was failing on 32-bit platforms due to `linenothreshold` defaulting to `sys.maxsize`.

Fixes: #522


chrisjsewell (approved):
cheers

PR #538, opened 2022-03-19
📚 DOCS: Fix typo - "to use write Sphinx"

bhagerty:
The phrase "allows you to use write Sphinx documentation" includes an extra word because of an editing oversight. I think the writer means "allows you to write Sphinx documentation" etc. My guess is that in a previous draft, the writer had "allows you to use MyST to write Sphinx documentation," and didn't fully complete the revision.


chrisjsewell (approved):
cheers

PR #541, opened 2022-04-01
🐛 FIX: Spurious newline in generated literal_block nodes

### LINKED ISSUE #407: Spurious newline in generated `literal_block` nodes
### Describe the problem

The result of `main.to_docutils` include an extra newline in each `literal_block` node. 

### Link to your repository or website

_No response_

### Steps to reproduce

Script:

```python
from myst_parser.main import to_docutils
from docutils.core import publish_doctree

doc = to_docutils("```\ncode\n```\n")
print(doc, end="\n\n")

ref = publish_doctree("::\n\n   code\n\n")
print(ref, end="\n\n")

from docutils.core import publish_from_doctree
print(publish_from_doctree(doc, writer_name="html4css1", settings_overrides={"embed_stylesheet": False}).decode("utf-8"), end="\n\n")
print(publish_from_doctree(ref, writer_name="html4css1", settings_overrides={"embed_stylesheet": False}).decode("utf-8"), end="\n\n")
```

Output:

```
<document source="notset"><literal_block language="" xml:space="preserve">code
</literal_block></document>

<document source="<string>"><literal_block xml:space="preserve">code</literal_block></document>

…
<body>
<div class="document">


<pre class="literal-block">
code

</pre>
</div>
</body>
</html>

…
<body>
<div class="document">


<pre class="literal-block">
code
</pre>
</div>
</body>
</html>
```

Notice the extra newlines in the output of MyST (compared to the expected docutils output `ref`)

### The version of Python you're using

3.8

### Your operating system

-

### Versions of your packages

_No response_

### Additional context

_No response_

cpitclaudel:
Closes #407.  The details are in the original bug report, but the long story short is that Docutils already adds a newline to its <pre> blocks, so we don't need one more.  I opted to fix all `literal_blocks`, but the most important ones are the ones in `create_highlighted_code_block`.


chrisjsewell:
Cheers @cpitclaudel, I will have a look probably Monday now

chrisjsewell:
Thanks @cpitclaudel this requires fixes for a number of the regression tests (if you didn't know you can run `pytest --force-regen`

My question would be though; should the behaviour be the same for docutils and sphinx? i.e.does the sphinx builder also require this new line removal?

PR #584, opened 2022-06-13
Add support for `hard_wrap`

rsuu:
Insert HTML `<br>` tags inside on paragraphs where the origin Markdown document had newlines

default: `myst_hard_wrap = False`
```html
<p>foo
bar</p>
```

`myst_hard_wrap = True`
```html
<p>foo<br>
bar</p>
```

p.s. I am noob in python

### DIFF
--- a/myst_parser/config/main.py
+++ b/myst_parser/config/main.py
@@ -75,11 +75,20 @@ class MdParserConfig:
             "help": "Use strict CommonMark parser",
         },
     )
+
     gfm_only: bool = dc.field(
         default=False,
         metadata={
             "validator": instance_of(bool),
             "help": "Use strict Github Flavoured Markdown parser",
+        },
+    )
+
+    hard_wrap: bool = dc.field(
+        default=False,
+        metadata={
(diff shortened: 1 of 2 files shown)


chrisjsewell:
Heya, you know hard wraps are already denoted in commonmark by backslashes `\` or two+ spaces? https://spec.commonmark.org/0.30/#hard-line-breaks

rsuu:
@chrisjsewell 

Yes I know.
But I want MyST can output `<p>foo<br>bar</p>` when I input `foo\nbar`.

This is an option in kramdown also.
> The hard_wrap option is only used by the GFM parser, not the standard kramdown parser. I will update the options page so that this is made clearer.
https://github.com/gettalong/kramdown/issues/217#issuecomment-73883813

p.s. Maybe replacing `myst_hard_wrap` with `myst_gfm_hard_wrap` will be better?


And I like to write this way:
```
aaa
bbb
ccc
```

Not this:
```
aaa\
bbb\
ccc\
```

Issue #591, opened 2022-06-26
Please test with Docutils 0.19b1


AA-Turner:
Hi,

We've just released Docutils 0.19b1, and intend to release 0.19 final on 05/07/2022 (a week on Tuesday). Please would you test with the pre-release, and raise any issues to me either on this issue or on the Docutils tracker?

- The release notes are at: https://docutils.sourceforge.io/RELEASE-NOTES.html#release-0-19b1-2022-06-21
- The detailed change history is at: https://docutils.sourceforge.io/HISTORY.html#release-0-19b1-2022-06-21
- You can install 0.19b1 with `python -m pip install --pre "docutils==0.19b1"` or install Docutils' master with `python -m pip install "git+https://repo.or.cz/docutils.git#subdirectory=docutils"`.

Thanks,
Adam

chrisjsewell:
Cheers for the notification @AA-Turner !

flokli:
I ran this with docutils 0.19, and got the following test failures:

```
=================================== FAILURES ===================================
____________________ test_docutils_roles[90-pep-reference] _____________________

file_params = ParamTestData(line=90, title='pep-reference', description='(`docutils.parsers.rst.roles.pep_reference_role`):', conten...g/dev/peps/pep-0000">\n            PEP 0\n', index=8, fmt=<pytest_param_files.main.DotFormat object at 0x7ffff4fd6ad0>)
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7ffff3ed8dc0>

    @pytest.mark.param_file(FIXTURE_PATH / "docutil_roles.md")
    def test_docutils_roles(file_params, monkeypatch):
        """Test conversion of Markdown to docutils AST (before transforms are applied)."""
    
        def _apply_transforms(self):
            pass
    
        monkeypatch.setattr(Publisher, "apply_transforms", _apply_transforms)
    
        doctree = publish_doctree(
            file_params.content,
            source_path="notset",
            parser=Parser(),
        )
    
>       file_params.assert_expected(doctree.pformat(), rstrip_lines=True)
E       AssertionError: Actual does not match expected
E       --- /build/source/tests/test_renderers/fixtures/docutil_roles.md:90
E       +++ (actual)
E       @@ -1,4 +1,4 @@
E        <document source="notset">
E       
E            <paragraph>
E       
E       -        <reference refuri="http://www.python.org/dev/peps/pep-0000">
E       
[...]

PR #606, opened 2022-08-04
🐛 FIX: don't fail with docutils 0.19

flokli:
Update the version bounds to allow docutils 0.19, and update the
fixtures to match the output produced by 0.19.

Fixes https://github.com/executablebooks/MyST-Parser/issues/591.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ keywords = [
 ]
 requires-python = ">=3.7"
 dependencies = [
-    "docutils>=0.15,<0.19",
+    "docutils>=0.19,<0.20",
     "jinja2", # required for substitutions, but let sphinx choose version
     "markdown-it-py>=1.0.0,<3.0.0",
     "mdit-py-plugins~=0.3.0",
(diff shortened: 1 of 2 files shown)


flokli:
As written in https://github.com/executablebooks/MyST-Parser/issues/591#issuecomment-1204861447, this now bumps the required version to 0.19. I'm not sure how to conditionally update the fixture contents to the output docutils 0.19 presents, only when docutils >0.19 is available.

I've seen a similar PR at https://github.com/executablebooks/MyST-Parser/pull/478 that patches the fixtures, but it looks like it's not patched conditionally. Did the tests with older versions just fail?

ssbarnea:
@flokli Any chance you could bring this PR to a form that could be merged? ... as in passing CI.

I am sure lots of us want to upgrade docutils, even if it might prove to be better to find a solution that does not drop support for older versions, as it is likely to cause impossible to address dependency conflicts with other docutils consumers.

flokli:
This PR was mostly intended to show what has changed with later versions of docutils. I still need some feedback from maintainers on how to proceed with this, as written in https://github.com/executablebooks/MyST-Parser/pull/606#issuecomment-1204873961.

chrisjsewell:
superceded by #611

PR #616, opened 2022-09-14
Update MySTyc URL

astrojuanlu:
The service does no longer need a backend, see https://github.com/astrojuanlu/mystyc/pull/2 🎉


chrisjsewell:
Wahoo, nice 🎉

Yeh was starting to also have a look at making one of these, that runs the myst-docutils python API, to convert MyST Markdown to HTML (or I guess any output format)

chrisjsewell (approved):

Issue #623, opened 2022-10-02
unnecessary translable string in sphinx's pot file


drafish:
### Describe the bug

**context**
I'v been working on remix docs which use `sphinx` to build the docs and used to use `recommonmark` as markdown parser.  There was an issue, and I'v already fixed it with `myst`. But it brings a new issue.

**expectation**
```
![](images/a-remixd-fe.png)
```
As you can see, this image takes the entire line. If I use `recommonmark`, `sphinx` will not identity it as a translable string. If the image has alt, `sphinx` will only extract alt as translable string. This is what I expected.

**bug**
But when I use `myst`, funny thing happened. For example:

In markdown, it was like this
```
![remixd](images/a-remixd-fe.png)
```
And this is what I got in pot file
```
![remixd](images/a-remixd-fe.png)

remixd
```
See, I got two translable string for one image. In order to fix this duplicate problem, I use this configuration (`myst_disable_syntax = ['image']`) to remove alt string. But it's just a temporary solution. What I really want is remain the alt string instead of the entire image string, and ignore all image string which does not have alt.

### Reproduce the bug

1. get a docs project which builded by `sphinx` and parsed by `myst`.
2. write some markdown like the bug description above, or you can just copy it.
3. use sphinx's gettext command to generate pot file from markdown

Or maybe you can just see this [pr of remix docs](https://github.com/ethereum/remix-ide/pull/3118). This kind of problems are all over the docs.

![image](https://user-images.githubusercontent.com/18543527/193436242-90b7aaf4-6151-444b-922f-43f645357257.png)

drafish:
Well, I kind of change my mind already. Maybe we should keep the entire image string, in case we need to replace image to other language version. What you think, Rob @ryestew

PR #624, opened 2022-10-02
FIX: unnecessary rawsource

drafish:
fix issue #623

### DIFF
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -439,7 +439,13 @@ class DocutilsRenderer(RendererProtocol):
     # ### render methods for commonmark tokens
 
     def render_paragraph(self, token: SyntaxTreeNode) -> None:
-        para = nodes.paragraph(token.children[0].content if token.children else "")
+        if (token.children):
+            if (token.children[0].children and len(token.children[0].children) == 1):
+                para = nodes.paragraph("")
+            else:
+                para = nodes.paragraph(token.children[0].content)
+        else:
+            para = nodes.paragraph("")
         self.add_line_and_source_path(para, token)
         with self.current_node_context(para, append=True):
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Thanks @drafish I will try to have a look at this soon; its just not clear to me if this is a "complete" solution for dealing with paragraph raw text.
Also obviously the test need to be fixed (hint you can run `pytest --force-regen` to auto-update https://pytest-regressions.readthedocs.io/ fixtures)

Perhaps also @jpmckinney wants to have a look, as the resident translation expert 😄

myst_parser/mdit_to_docutils/base.py
```diff
@@ -439,7 +439,13 @@ def renderInlineAsText(self, tokens: list[SyntaxTreeNode]) -> str:
     # ### render methods for commonmark tokens
 
     def render_paragraph(self, token: SyntaxTreeNode) -> None:
-        para = nodes.paragraph(token.children[0].content if token.children else "")
+        if token.children:
+            if token.children[0].children and len(token.children[0].children) == 1:
```
jpmckinney, line 443: Can you add a test case for this scenario? (the image markup example from the issue description)

jpmckinney: I'm also not sure that this is the correct condition. I assume it works to correct the specific issue with the image markup, but wouldn't this also mean that any paragraph with a single child node ends up with no content?

drafish:
I change my mind. I always thought image does not need to be translated. But that's not right. Not only alt can be translated, image can also be translated. MyST is right, recommonmark is wrong.

Even if my thought is right, my code is wrong. I may have tested it in a wrong way. Now I tested it again, my code can not render image correctly. 

I am sorry for my careless work. I should trust MyST more. I am closing this pr now. @chrisjsewell @jpmckinney

PR #635, opened 2022-10-19
🧪 Introduce a gate/check GHA job

webknjaz:
This adds a GHA job that reliably determines if all the required dependencies have succeeded or not.

It also allows to reduce the list of required branch protection CI statuses to just one — `check`. This reduces the maintenance burden by a lot and have been battle-tested across a small bunch of projects in its action form and in-house implementations of other people.

This action is now in use in aiohttp (and other aio-libs projects), CherryPy, conda, coveragepy, Open edX, Towncrier some of the Ansible repositories, pip-tools, all of the jaraco's projects (like `setuptools`, `importlib_metadata`), some of hynek's projects (like `attrs`, `structlog`), some PyCQA, PyCA, PyPA and pytest projects, a few AWS Labs projects. Admittedly, I maintain a few of these but it seems to address some of the pain folks have:
https://github.com/jaraco/skeleton/pull/55#issuecomment-1106638475.

I figured, this might be useful for MyST too, which is why I'm submitting this patch.

The story behind this is explained in more detail at https://github.com/marketplace/actions/alls-green#why.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,12 @@ jobs:
           python-version: "3.8"
           sphinx: ">=5,<6"
 
+    continue-on-error: >-
+      ${{
+        (matrix.python-version == '3.8' && matrix.sphinx == '>=5,<6')
+        && true || false
+      }}
+
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -36,6 +36,12 @@ jobs:
           python-version: "3.8"
           sphinx: ">=4,<5"
 
+    continue-on-error: >-
```
webknjaz, line 39: @chrisjsewell this is here to match the currently set up branch protection.

.github/workflows/tests.yml
```diff
@@ -103,10 +109,29 @@ jobs:
     - name: Run docutils CLI
       run: echo "test" | myst-docutils-html
 
+  # https://github.com/marketplace/actions/alls-green#why
+  check:  # This job does nothing and is only used for the branch protection
```
webknjaz, line 113: This is what should eventually replace all the individual branch protection checks that belong to this workflow.

.github/workflows/tests.yml
```diff
@@ -103,10 +109,29 @@ jobs:
...
+
+    steps:
+    - name: Decide whether the needed jobs succeeded or failed
+      uses: re-actors/alls-green@release/v1
+      with:
+        jobs: ${{ toJSON(needs) }}
+
   publish:
 
     name: Publish myst-parser to PyPi
```
webknjaz, line 132: An unrelated comment/suggestion for the future: this could use an `environment` setting and secrets could be scoped to it instead of being accessible globally.

.github/workflows/tests.yml
```diff
@@ -36,6 +36,12 @@ jobs:
           python-version: "3.8"
           sphinx: ">=5,<6"
 
+    continue-on-error: >-
+      ${{
+        (matrix.python-version == '3.8' && matrix.sphinx == '>=5,<6')
+        && true || false
+      }}
+
```
chrisjsewell, line 44:
```suggestion
```

.github/workflows/tests.yml
chrisjsewell: I removed this as I think its not necessary to ignore any

chrisjsewell:
Cheers!

PR #669, opened 2023-01-06
ci(deps): setup dependabot 

mmorel-35:
Provides dependabot configuration to check weekly if there is a new version available. It creates or replace a pr per dependency. I defined a prefix as I saw that you are using the up arrow for dependency upgrades.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


chrisjsewell (approved):
cheers

PR #678, opened 2023-01-10
👌 IMPROVE: Allow for heading anchor links in docutils

chrisjsewell:
This aligns the treatment of `[](#target)` style links for docutils with sphinx, such that they are linked to a heading slug.

The core behaviour for sphinx is not changed,
except that failed reference resolution
now emits a `myst.xref_missing` warning (as opposed to a `std.ref` one), with a clearer warning message. Also on failure, the reference is still created,
for people who wish to suppress the warning (see e.g. #677)

This is another step towards https://github.com/executablebooks/myst-enhancement-proposals/pull/10


chrisjsewell:
@choldgraf @rowanc1 please review this asap; it is the final commit before a new version can be released.

Hopefully you can see from the recent commits, the point of these is to facilitate moving towards an implementation for https://github.com/executablebooks/myst-enhancement-proposals/pull/10; implementing the things that are not yet breaking backwards compatibility.

Obviously, that PR is essentially useless if what it stipulates can't actually be implemented, in a robust way

docs/syntax/syntax.md
```diff
@@ -255,7 +255,8 @@ By default, MyST will resolve link destinations according to the following rules
 3. Destinations which point to a local file path are treated as links to that file.
    - The path must be relative and in [POSIX format](https://en.wikipedia.org/wiki/Path_(computing)#POSIX_and_Unix_paths) (i.e. `/` separators).
    - If the path is to another source file in the project (e.g. a `.md` or `.rst` file),
-    then the link will be to the initial heading in that file.
+    then the link will be to the initial heading in that file or,
+    if the path is appended by a `#target`, to the heading "slug" in that file.
```
choldgraf, line 259: What happens if `#target` is also defined as a section label as well? E.g. if there's content like:

```
(target)=
## Some section

Some text

## Target

Some other text, and a ref to [](#target).
```

And does the behavior change at all if instead the reference was `[](myfile.md#target)`?

choldgraf (approved):
This looks good to me - I think the behavior you describe here is what others expect in markdown link behavior anyway. One quick question but I don't think it needs to block the PR if you want to get a release out quickly.

chrisjsewell:
> This looks good to me

thanks!

docs/syntax/syntax.md
choldgraf: ah ok - I thought this was implementing that part of the MEP, but if this is just a step forward in _anticipation_ of the MEP, I think we should just merge it as-is

chrisjsewell: > I thought this was implementing that part of the MEP,

yeh, no because obviously that could start breaking peoples projects, so that needs to be done with great care 😅 (and ideally provide them a way to migrate) 

With https://github.com/executablebooks/myst-enhancement-proposals/pull/10, we would like `#target` to search through essentially all targets on that page (and possibly all pages); headings, figures, tables, equations, ...

1. You need a systematic way to do this
2. and you need to allow for a way to filter on ambiguous target names, i.e. by specifying the domain and/or object type (similar to the new `inv:key:domain:type#name` links)

In #613 I have kind of achieved this, but it does involve some "patching" of sphinx's code, for example they do some weird stuff with creating latex labels, which I guess does introduce a bit more potential maintenance burden.

PR #679, opened 2023-01-11
📚 DOCS: Add live preview (w/ pyscript)

chrisjsewell:
Using https://docs.pyscript.net

See https://myst-parser--679.org.readthedocs.build/en/679/live-preview.html 🚀


chrisjsewell:
Another one for you @choldgraf

rowanc1:
Pretty cool to be able to have live docs on everything. Love seeing the docutils ast in here as well, very helpful.

chrisjsewell:
> Love seeing the docutils ast in here as well, very helpful.

yeh obviously be nice to eventually sync up with the "myst AST"
(to note, this AST is a representation after transforms)

chrisjsewell:
I quite like that https://github.com/pyscript/pyscript/tree/main/docs uses myst-parser, and now myst-parser uses pyscript, and nice 🔃

docs/live-preview.md
```diff
@@ -0,0 +1,96 @@
...
+
+:::::{tab-set}
+::::{tab-item} Input text
+````{raw} html
+<textarea class="pyscript" id="input_myst">
+# Heading 1
+
+Hallo world!
+
+term
```
choldgraf, line 37:
```suggestion
```{note}
An admonition note!
```

term
```

**Review by choldgraf — COMMENTED** [review]
This is super cool - a few quick thoughts:

1. What do you think about putting a `Live Demo` button just to the right of `Get Started`?
2. When the pyodide environment is loading, there's a weird UI thing that goes on where most of the page is greyed out, except for the article header. Is that a quick fix? Here's a screenshot:

    image

File: docs/live-preview.md [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+    - docutils==0.19
+    - pygments
+  splashscreen:
+    autoclose: true
+---
+
+# Live Preview
+
+This is a live preview of the MyST Markdown [docutils renderer](docutils.md).
+You can edit the text/configuration below and see the live output.
```
choldgraf, line 14:
```suggestion
You can edit the text/configuration below and see the live output.
Note that there is no styling applied (that is usually provided by Sphinx themes).
```

chrisjsewell:
> What do you think about putting a Live Demo button just to the right of Get Started?

done 👍 

>  Is that a quick fix? 

yeh dunno how to change that, will leave for now

docs/live-preview.md
chrisjsewell: FYI, you can specify the CSS, I changed it in the config from the built-in default `plain.css` to `minimal.css`, which is a little nicer 😄

PR #682, opened 2023-01-12
👌 Change non-fatal directive parsing errors to warnings

chrisjsewell:
For non-fatal errors, such as;
faulty options syntax, unknown option keys, and invalid option values,
a warning is raised, but the directive is still run (without the erroneous options).
The warning is given the `myst.directive_parse` type, which can be suppressed.


rowanc1:
Very much want to see this in the JS side as well, it is a bit too strict.

rowanc1:
I only looked at the tests, but this seems like a great addition to me 👍. I will follow suit in the JS side soon!

chrisjsewell:
> Very much want to see this in the JS side as well, it is a bit too strict.

yep exactly, I know we talked about this

Either in this PR or a follow, was going to add changing unknown directives and roles to a warning, with a certain warning type

rowanc1:
Awesome, can you tag me so I can make the JS one the same!

choldgraf:
> was going to add changing unknown directives and roles to a warning, with a certain warning type

+1 from me

PR #684, opened 2023-01-13
📚 DOCS: Add the sphinx-tippy extension

chrisjsewell:
Inspired by @rowanc1, I created https://sphinx-tippy.readthedocs.io 😄

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,7 @@
 version: 2
 
 python:
-  version: "3"
+  version: "3.8"
   install:
       - method: pip
         path: .
--- a/docs/_static/local.css
+++ b/docs/_static/local.css
@@ -1,16 +1,16 @@
 /** Add a counter before subsections **/
-h1 {
+h1:not(.tippy-header) {
     counter-reset: subsection;
     text-decoration: underline;
 }
-h2 {
+h2:not(.tippy-header) {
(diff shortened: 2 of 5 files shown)


rowanc1:
Oh wow, that is super cool to see this functionality making it's way into sphinx. Are you doing all/most of this at compile time?

chrisjsewell:
>  Are you doing all/most of this at compile time?

yeh I've implemented it to do it all at compile time, so I don't have to deal with CORS, caching, etc 😅 

although, as mentioned https://sphinx-tippy.readthedocs.io/en/latest/#how-does-it-work, there is also https://sphinx-hoverxref.readthedocs.io/ that RTD guys run that requires you be running the website with them hosting

rowanc1:
Yeah, makes sense. Certainly difficult to do without a service running.

chrisjsewell:
yeh was just interested to see how/what could be done statically

chrisjsewell:
Ah RTD also has a content API: https://readthedocs.org/api/v3/embed/?url=https://myst-parser.readthedocs.io/en/latest/intro.html%23get-started

chrisjsewell:
> Ah RTD also has a content API

@rowanc1 I guess this type of content API, returning HTML, isn't ideal though, because you are then going to apply your local CSS which might not be "compatible"

PR #686, opened 2023-01-15
👌 Add `myst_enable_checkboxes` config

chrisjsewell:
@asmeurer is this what you are after?


chrisjsewell:
hmm not working yet

asmeurer:
Yes, this works. This is much better than the manual `<li><input type="checkbox">` I was doing previously.

I would still like the other point in the issue (about removing the bullets from the checkboxes), but I can workaround that with some custom CSS for now.

PR #691, opened 2023-01-27
📚 Add html_last_updated_fmt = "" to conf.py

jeanas:
Should prevent "Last updated on None" in the footer.


chrisjsewell:
It does indeed, thanks!

PR #695, opened 2023-02-10
✨ NEW: extended URL link customisation

chrisjsewell:
The `myst_url_schemes` configuration has been extended, in a back-compatible manner,
to also allow writing as a dictionary, with customisations of the link URL/text, via templates, such as:

```python
myst_url_schemes = {
    "http": None,
	"https": None,
    "wiki": "https://en.wikipedia.org/wiki/{{path}}#{{fragment}}",
    "gh-issue": {
        "url": "https://github.com/executablebooks/MyST-Parser/issue/{{path}}#{{fragment}}",
        "title": "Issue #{{path}}",
        "classes": ["github"],
    },
}
```

Allowing for `<gh-issue:639>` and `[URI](wiki:Uniform_Resource_Identifier#URI_references)`.

The `inline_attrs` extension also allows for specific links to be marked as external, using `[](my-link){.external}`.


chrisjsewell:
cc @choldgraf @rowanc1 

See: https://myst-parser--695.org.readthedocs.build/en/695/syntax/syntax.html#customising-external-url-resolution

This is essentially a more extensible version of https://myst-tools.org/docs/mystjs/external-references#wikipedia-links etc

rowanc1:
Only looked through the docs and not the code, this looks great. On the face of it very similar to the like transformers we have in mystjs, but allows these to be set as data. The config options look sensible.

There is also very similar functionality in the extlink in sphinx:
https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html#confval-extlinks

How do you think this compares to extlink, and/or the advantages of this to address similar/same functionality? I like that these are picked up as markdown links, but maybe the config can be closer, or the same as sphinx already has?

chrisjsewell:
> How do you think this compares to extlink, and/or the advantages of this to address similar/same functionality?

These allow word functionality, by not simply using the `%s`, which to some degree is a deprecated aspect of python (replace by f-strings), by allowing the url to be fully decomposed. The dict format also allows future extensibility (for mor keys), rather than the tulple format, and this is already apparent in the ability to add classes to specific schemes.

Note I explicitly don't allow full jinja templating or actual python function manipulation of the string, in order to still allow for a "language-agnostic" configuration. I believe this is a good middle ground, of "full-extensibility" vs cross-language compatibility

rowanc1:
Yep makes sense, the tuple format isn't very customizable, and this allows for named parts but stays close to a familiar format (jinja) without the complexity (which isn't necessary for links).

I feel like this is something that we could pick up really easily in mystjs from a config/functionality standpoint so don't see any divergence there!

I like the idea of having some of these turned on by default in jupyterbook in the future, especially the wiki/doi links.

PR #696, opened 2023-02-10
👌 IMPROVE: Allow heading_slug_func to be a string 

chrisjsewell:
`myst_heading_slug_func` can now also be set to a string,
which will be interpreted as an import path to a function,
e.g. `myst_heading_slug_func = "mypackage.mymodule.slugify"`.

### DIFF
--- a/docs/syntax/optional.md
+++ b/docs/syntax/optional.md
@@ -520,6 +520,12 @@ The anchor "slugs" created aim to follow the [GitHub implementation](https://git
 
 To change the slug generation function, set `myst_heading_slug_func` in your `conf.py` to a function that accepts a string and returns a string.
 
+:::{versionadded} 0.19.0
+`myst_heading_slug_func` can now also be set to a string,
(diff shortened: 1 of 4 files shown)


docs/syntax/optional.md
```diff
@@ -520,6 +520,12 @@ The anchor "slugs" created aim to follow the [GitHub implementation](https://git
 
 To change the slug generation function, set `myst_heading_slug_func` in your `conf.py` to a function that accepts a string and returns a string.
 
+:::{versionadded} 0.19.0
+`myst_heading_slug_func` can now also be set to a string,
+which will be interpreted as an import path to a function,
+e.g. `myst_heading_slug_func = "mypackage.mymodule.slugify"`.
+:::
```
Alexey-NM, line 527: Greate change, thank you.
I think it would be better to add an advice. Something like:
>You should set this parameter as string to avoid unnecessary rebuild project. 

It will save a lot of time for other people.

PR #703, opened 2023-02-18
✨ NEW: Add `attrs_block` extension

chrisjsewell:
By adding `"attrs_block"` to `myst_enable_extensions` (in the sphinx `conf.py`), you can enable parsing of inline attributes after certain inline syntaxes. This is adapted from [djot block attributes](https://htmlpreview.github.io/?https://github.com/jgm/djot/blob/master/doc/syntax.html#block-attributes)


chrisjsewell:
cc @mgielda  https://myst-parser--703.org.readthedocs.build/en/703/syntax/optional.html#block-attributes

PR #707, opened 2023-02-20
👌 Enhance `[](#id)` references

chrisjsewell:
This PR moves myst-parser in the direction of https://github.com/executablebooks/myst-enhancement-proposals/pull/10, in a relatively back compatible manner, and in a way that supports both docutils and sphinx.

It expands the capability of `[](#id)` to more than just linking to heading slugs, with the order of specificity being:

1. If it matches a local (to that document) "explicit" `std:ref` target, then link to that and stop
   - Note, currently only `std:ref` domain/type are supported, e.g. not `math` etc. That's more difficult and can come later
2. If it matches a local (to that document) "implicit" heading slug, then link to that and stop
3. If using docutils (i.e. single-page) build, then stop here and emit a `myst.xref_missing` warning
4. If using sphinx then create a `pending_xref` node, and hand-off to sphinx's "any" resolver, which takes effect once all documents have been read:
    - This first tries to resolve against a local (to the project) reference and, if matching, stops
    - Otherwise also try to match against any intersphinx reference
    - Otherwise emit a ~~`myst.ref`~~ `myst.xref_missing`  warning
 
If the text is explicit, e.g. `[text](#id)`, that text is used, otherwise a determination of implicit text is attempted, e.g. based on the section title or figure caption.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,7 @@ coverage:
   status:
     project:
       default:
-        target: 90%
+        target: 89%
         threshold: 0.5%
     patch:
       default:
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -248,36 +248,14 @@ class DocutilsRenderer(RendererProtocol):
     def _render_finalise(self) -> None:
         """Finalise the render of the document."""
 
-        # attempt to replace id_link references with internal links
-        for refnode in findall(self.document)(nodes.reference):
-            if not refnode.get("id_link"):
-                continue
-            target = refnode["refuri"][1:]
-            if target in self._slug_to_section:
-                section_node = self._slug_to_section[target]
-                refnode["refid"] = section_node["ids"][0]
-
-                if not refnode.children:
-                    implicit_text = clean_astext(section_node[0])
-                    refnode += nodes.inline(
-                        implicit_text, implicit_text, classes=["std", "std-ref"]
-                    )
-            else:
-                self.create_warning(
-                    f"local id not found: {refnode['refuri']!r}",
-                    MystWarnings.XREF_MISSING,
-                    line=refnode.line,
-                    append_to=refnode,
-                )
-                refnode["refid"] = target
-            del refnode["refuri"]
-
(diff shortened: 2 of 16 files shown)


chrisjsewell:
@rowanc1 I just got to fix/add a few tests for this, but can you (hopefully) agree in principle this is all good cheers

chrisjsewell:
Ok this is good to go, so will be merging relatively soon, so I can get a release out finally

rowanc1:
From the description this looks great, looked through a few of the tests -- all makes sense!

This looks to address `Review "search order specificity" and "implicit" targets in a document` in the [cross-reference MEP](https://github.com/executablebooks/myst-enhancement-proposals/pull/10)?

Is this the final piece that you think is addressing backwards compatibility with the references MEP?

chrisjsewell:
> From the description this looks great, looked through a few of the tests -- all makes sense!

Thanks!

> Is this the final piece that you think is addressing backwards compatibility with the references MEP?

Its getting there 😅 

For example, it doesn't address things like the ``{name}`` ``{number}`` placeholders.
Obviously, just trying to eat away at the "easier" aspects, and yeh, trying to do it in a way that's

- relatively back-compatible
- works where possible for both docutils and sphinx
- doesn't override too much sphinx code (that could be liable to break in later versions)

rowanc1:
Should that be another place we pair back the MEP for now, especially as the enumeration isn't going to work for `objects.inv` for a while/ever. Or we can put notices on those parts of the MEP, as they are where we are planning to go, but it might take a while.

chrisjsewell:
> Should that be another place we pair back the MEP for now,

yeh I guess quite possibly. it would be a shame to break up the "holistic" goal of the MEP.

But perhaps we do something like keep a record of the "full one" and progress with a "lite version"

rowanc1:
Yeah, I think we should probably pare back and get these other ideas on a separate draft MEP right away to keep track of our intent.

I think it is removing/amending:
- inv/myst
- enumeration and name replacement

Anything else?

chrisjsewell:
> Anything else?

I think they are probably the main things 👍

chrisjsewell:
ok merging

PR #711, opened 2023-02-21
👌 Handling of nested headers

chrisjsewell:
For the longest time, nested headers in myst-parser have been a pain, particularly in things like admonitions

In Markdown (and HTML) headings are allowed "anywhere", for example:

```markdown
> # Heading 1
## Heading 2

Paragraph
```

Is rendered as:

```html
<blockquote>
    <h1>Heading 1</h1>
</blockquote>
<h2>Heading 2</h2>
<p>Paragraph</p>
```

However, because docutils/sphinx treats headers as nested sections, this becomes problematic

```xml
<blockquote>
   <section>
		<title>
            Heading 1
<section>
   <title>
      Heading 2
   <paragraph>
      Paragraph
```

Which sphinx cannot resolve the ToC tree from etc

This PR fixes this, by identifying if a heading is inside another component and instead outputting it as a "non-structural" rubric node

```xml
<blockquote>
   <rubric level=1>
		Heading 1
<section>
    <title>
      Heading 2
   <paragraph>
      Paragraph
```

Natively, docutils/sphinx does not deal with the "level" key in the rubric,
so here we also override the rubric HTML renderer to correctly output a `<h>` element, if "level" is present, to retrieve the desired:

```html
<blockquote>
    <h1>Heading 1</h1>
</blockquote>
<h2>Heading 2</h2>
<p>Paragraph</p>
```

~~Note, when such an identification is made, a `myst.nested_header` warning is emitted:~~

There is no longer any warning of nested headers, since this is the intended behaviour

~~Additionally, using the new `attrs_block` syntax extension, a user can specifically specify that a heading should not be included in the ToC~~

This was removed

### DIFF
--- a/docs/syntax/optional.md
+++ b/docs/syntax/optional.md
@@ -859,13 +859,6 @@ For **block quotes**, the `attribution` key is supported:
 {attribution="Chris Sewell"}
 > Hallo
 
-For **headings**, the `toc` key marks the heading as not being included in the table of contents:
-
-:::{myst-example}
-{toc=false}
-# My Heading
-:::
-
 (syntax/attributes/inline)=
 ### Inline attributes
 
--- a/docs/syntax/optional.md
+++ b/docs/syntax/optional.md
@@ -859,6 +859,13 @@ For **block quotes**, the `attribution` key is supported:
 {attribution="Chris Sewell"}
 > Hallo
 
+For **headings**, the `toc` key marks the heading as not being included in the table of contents:
+
+:::{myst-example}
+{toc=false}
+# My Heading
+:::
+
 (syntax/attributes/inline)=
 ### Inline attributes
 
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -769,29 +769,34 @@ class DocutilsRenderer(RendererProtocol):
     def render_heading(self, token: SyntaxTreeNode) -> None:
         """Render a heading, e.g. `# Heading`."""
 
+        level = int(token.tag[1])
+
         if (
-            token.attrs.get("toc", None) == "false"
+            # the heading is not at the root level
+            (token.parent and token.parent.type != "root")
+            # the heading is specifically markded as not being included in the ToC
+            or token.attrs.get("toc", None) == "false"
+            # the heading is being passed within a nested parse
+            # which has not been specifically marked as allowing nested headers
             or self.md_env.get("match_titles", None) is False
         ):
+            # since this would break the document structure
+            # we instead create a rubric node, which is a "non-structural" heading
+
             if token.attrs.get("toc", None) != "false":
-                # this can occur if a nested parse is performed by a directive
-                # (such as an admonition) which contains a header.
-                # this would break the document structure
                 self.create_warning(
-                    "Disallowed nested header found, converting to rubric",
+                    "Nested header will not be included in the ToC",
                     MystWarnings.MD_HEADING_NESTED,
                     line=token_line(token, default=0),
                     append_to=self.current_node,
                 )
 
-            rubric = nodes.rubric(token.content, "")
+            rubric = nodes.rubric(token.content, "", level=level)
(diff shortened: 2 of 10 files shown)


chrisjsewell:
cc @choldgraf @rowanc1 I'm pretty pleased with this 😄 

Note, For the `{toc=false}` block attribute syntax, I'm discussing "best practice" here: https://github.com/jgm/djot/issues/213

chrisjsewell:
Actually, scrap the additional attribute logic, I have a better solution, and a few changes to make here...

chrisjsewell:
Ok updated and ready to go!

To clarify, the logic is now:

- A `section` can only be a child of the root `document`, or another `section`
- If a header token is encountered, with a child that is not one of these, then it is added as a `rubric`
- Otherwise a new `section` is created, and the heading is added a `title`, which is a child of the `section`

(This is essentially exactly the same logic as djot 😄 )

There is no longer any warning about nested headings, because it is the intended behaviour of MyST (and should be documented as such)

One exception to this rule 😒  is that a few special directives (such as `only`) can specify that sections should be allowed to be created as a child of the current node, e.g.

```python
def MyDirective(Directive):
	def run(self) -> list[nodes.Node]:
         node = nodes.Element()
         self.state.nested_parse(self.content, self.content_offset, node, match_titles=True)
         return node.children
```

This is handled by the changes in this PR

chrisjsewell:
Does this make sense @rowanc1 😄

choldgraf:
From a UX perspective this makes total sense to me 👍

rowanc1:
As a note: in JupyterLab they use `tocSkip` (see [here](https://github.com/jupyterlab/jupyterlab/blob/bf2889c1571ba62426bfd0ed23976a6ce89b839f/packages/toc/src/utils/markdown.ts#L246)) which might be good to standardize on.

The section nesting is something we are also doing in the JATS export, which requires more structure (similar to RST). This has been somewhat at odds with the block structure (e.g. do blocks/cells define structure or do headings?!), but hasn't been too much of a problem so far, and we can at least transform between blocks --> sections very easily.

Overall, excited to see the nested heading errors go away!

chrisjsewell:
> This has been somewhat at odds with the block structure (e.g. do blocks/cells define structure or do headings?!)

Oh yeh I know that one is a pain 😅

chrisjsewell:
Ok so it looks like we are in agreement here, that I can merge this?

chrisjsewell:
@rowanc1 another thing I want to do after this is merged, as a result of https://github.com/jgm/djot/issues/213, is change a bit how `colon_fences` work, to be more like djot divs
Since you are using them for admonitions, I imagine you would like to see them in myst core syntax 😄 

So at present, in myst-parser, this is treated as an admonition directive 👍 

```
:::{note}
Some text
:::
```

```xml
<note>
   <paragraph>
      Some text
```

But, this is treated like a code block 👎 

```
:::name
Some text
:::
```

```xml
<code language=name>
    Some text
```

I think, in keeping with the purpose of these block (and djot), it should be converted to

```xml
<div class=name>
   <paragraph>
      Some text
```

does that make sense?

rowanc1:
Yes please to improved colon-fences. I see another MEP in our future. I like having more specificity on the colon blocks as to when to use them and why. I [sketched a bit here](https://myst-tools.org/docs/mystjs/syntax-overview#directives) about use colons when the content is markdown-like.

I would also like to be more like pandoc on the nesting front (i.e. a start of a div only requires three colons + attribute, and the end has any number of colons, but no attribute).

chrisjsewell:
> i.e. a start of a div only requires three colons + attribute, and the end has any number of colons, but no attribute

yep as per djot, i.e. pandoc v2 😝 

> A div begins with a line of three or more consecutive colons, optionally followed by white space and a class name (but nothing else). It ends with a line of consecutive colons at least as long as the opening fence, or with the end of the document or containing block.

> The contents of a div are interpreted as block-level content.

mmcky:
oh this is going to be really useful and will solve an issue we just had with titles nested in `exercise` nodes.

PR #713, opened 2023-02-21
👌 Improve: Non-directive colon-fences

chrisjsewell:
At present, colon fences without a directive specified (e.g. `:::{note}`) are treated as code bocks 

```
:::name
Some text
:::
```

```xml
<code language=name>
    Some text
```

In keeping with the purpose of these block, as containers for nested MyST content, they are now converted to:

```xml
<div class=name>
   <paragraph>
      Some text
```

This is also in keeping with the original pandoc/djot inspiration <https://htmlpreview.github.io/?https://github.com/jgm/djot/blob/master/doc/syntax.html#div>

### DIFF
--- a/docs/syntax/optional.md
+++ b/docs/syntax/optional.md
@@ -477,6 +477,7 @@ This text is **standard** _Markdown_
 This text is **standard** _Markdown_
 :::
 
+
 (syntax/admonitions)=
 
 ## Admonition directives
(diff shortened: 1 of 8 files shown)


chrisjsewell:
cc @rowanc1 @choldgraf

rowanc1:
Makes sense to me, and like that there is a difference between colons and backticks.

PR #717, opened 2023-02-23
📚 DOCS: Full update

chrisjsewell:
Just about ready for a release!


rowanc1 (dismissed):
Took a look through all the docs, block attrs and inline attributes look great. The hover-tool tips are great, as you know I am a big fan of showing contextual information. Also was funny to see some of the Shakespeare quote and poems I chose making their way back here.

Only concern in this PR is around the mixing of directive names and class names. I think that is fine if we adopt a widely used existing syntax (e.g. support quarto-style callouts). However, I don't think that we should suggest another way to do admonitions without addressing some of the existing confusions. e.g. in https://github.com/executablebooks/myst-spec/issues/49. @mmcky @choldgraf and I have been chatting about this recently, and some of our thoughts are going to be clarified in the next few weeks!

Can we please either (1) put a warning around this saying we might deprecate it; or (2) remove the functionality.

I would love to discuss how directives and class-names interact in a specific MEP, but I think that needs to come _after_ we decide as a team on the block and inline syntax!

docs/syntax/admonitions.md
```diff
@@ -0,0 +1,148 @@
...
+
+:::{myst-example}
+
+{.note}
+This is a note admonition.
+
+{.warning title="This is a *warning*!"}
+This is a warning admonition.
+
+:::
```
rowanc1, line 89: (didn't hit save, woops.)

Can we put a warning that this might change in the future please. I think we need a MEP to define how the block attributes and directive kinds interact.

My concern is that there are already many ways to create admonitions in various tools, I would rather not add another to the mix. For class-names, I could certainly see using the `warning` class in CSS, and being surprised that this syntax sometimes turns into an admonition.

chrisjsewell:
Just to link, as we discussed on slack; I have tried to kick start some conversation here, as to a shorthand to essentially specify a role/directive in attribute syntax: https://github.com/jgm/djot/issues/146

docs/syntax/admonitions.md
chrisjsewell: As per #719, I was indeed looking to cherry-pick this implementation out into a separate PR, to discuss 👍

docs/syntax/optional.md
```diff
@@ -991,15 +875,7 @@ The figure block must contain **only** two components; an image, in either Markd
...
-```md
-:::{figure-md} fig-target
-:class: myclass
-
-<img src="img/fun-fish.png" alt="fishy" class="bg-primary mb-1" width="200px">
-
-This is a caption in **Markdown**
-:::
-```
+::::{myst-example}
```
mgielda, line 878: I think the docs should probably use sth like:

```
![fishy](img/fun-fish.png)
```

For an example. Hopefully most people use the md image syntax, not the HTML one :)

docs/syntax/admonitions.md
chrisjsewell: yep I've removed this

docs/syntax/optional.md
chrisjsewell: Done thanks!

PR #732, opened 2023-03-01
📚 DOCS: Un-wrap announcement in <p>

choldgraf:
The announcement text at the top was having a base pydata theme color applied to it because it was wrapped in `<p>`, the simplest fix is just to un-wrap it for now, though this should probably be made more robust in the pydata theme as well.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -159,7 +159,7 @@ html_theme_options = {
     "use_repository_button": True,
     "use_edit_page_button": True,
     "use_issues_button": True,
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Ah ok cheers, I thought it looked a little off
I used the `<p>` because it was in the documentation: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html#provide-local-html-in-your-theme, so yeh might wanna look into that

choldgraf:
ooh good call:

- https://github.com/pydata/pydata-sphinx-theme/pull/1223

PR #738, opened 2023-03-06
📚 Fix the sphinx-design example

recfab:
Under the `## Extending Sphinx` heading, the text of the example discusses `sphinx-design`, but the pip install command was installing `sphinxcontrib.mermaid`. I see there is a commented-out example, below, which is related to mermaid, so perhaps this was missed in a copy/paste.

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -133,7 +133,7 @@ For example, let's install the [sphinx-design](https://github.com/executablebook
 First, install `sphinx-design`:
 
 ```shell
-pip install sphinxcontrib-mermaid
+pip install sphinx-design
 ```
 
 Next, add it to your list of extensions in `conf.py`:


chrisjsewell:
Yep cheers 🙏 

When mermaid gets fixed, then it can come back out of the shadows lol (https://github.com/mgaitan/sphinxcontrib-mermaid/issues/109)

recfab:
Oh! I just noticed that you merged it. Thanks!

Initially, I had created this PR in draft-mode currently so I could amend the commit message to conform to the guidelines listed at https://github.com/executablebooks/.github/blob/master/CONTRIBUTING.md#commit-messages. Is not that considered required in this case?

chrisjsewell:
> Is not that considered required in this case?

no need for the formalities, I cahnaged the PR/commit message to have my emojis and thats all. Thanks for the contribution!

recfab:
Glad to help. Thanks for the quick merge!

PR #753, opened 2023-03-17
👌 Improve default slug generation for heading anchors

### LINKED ISSUE #752: Difference with github in default slug generation
I think `.strip()` is wrong in `default_slugify()`:

https://github.com/executablebooks/MyST-Parser/blob/a784880c7fbd893566087b83deec081376740e2c/myst_parser/mdit_to_docutils/base.py#L1969-L1977

Have a look at [this gist testing github slugs](https://gist.github.com/Cimbali/201095c2ad6533adc0747ab274d804de), you can hover the links at the left-hand side to check the correct slugs.

If there is a leading/trailing object such as an image, the stripping happens before removing the image (as in myst-parser currently). When slugifying, the image is removed, but no subsequent `.strip()` happens, which means slugs have a leading/trailing `-`.

Just removing `.strip()` should fix the behaviour (but maybe you want to add tests somewhere?)

Cimbali:
This can be easily tested by visiting the source markdown page on github:

https://github.com/Cimbali/MyST-Parser/blob/master/tests/test_sphinx/sourcedirs/references/index.md#image-in-title-

Fixes #752.

----

Note that for the `nested $a=1$` case I’ve updated the test to match the new behaviour, to allow improving matching github behaviour bit by bit.

Neither the old or new test match Github behaviour:
https://github.com/executablebooks/MyST-Parser/blob/master/tests/test_sphinx/sourcedirs/references/index.md#title-with-nested-a1

The issue seems to be math-to-text conversion (so the trailing `-` added by this PR is “more correct” even though not 100% there).

### DIFF
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -1974,7 +1974,7 @@ def default_slugify(title: str) -> str:
     - https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
     - https://gist.github.com/asabaylus/3071099
     """
-    return _SLUGIFY_CLEAN_REGEX.sub("", title.strip().lower().replace(" ", "-"))
+    return _SLUGIFY_CLEAN_REGEX.sub("", title.lower().replace(" ", "-"))
 
 
 def compute_unique_slug(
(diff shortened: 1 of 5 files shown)


chrisjsewell:
Thanks @Cimbali! I'll have a proper look probably over the weekend

Cimbali:
Hi @chrisjsewell have you been able to have a look?

It’s quite simply keeping the trailing `-` in slugs, as you can see for example on the first page I linked.

PR #769, opened 2023-05-26
fix: Do not raise an AttributeError if env is not available

### LINKED ISSUE #768: SphinxRenderer.sphinx_env must try/except, like DocutilsRenderer.sphinx_env
### Describe the bug

**context**
When I do:

```python
from myst_parser.config.main import MdParserConfig
from myst_parser.mdit_to_docutils.base import make_document
from myst_parser.mdit_to_docutils.sphinx_ import SphinxRenderer
from myst_parser.parsers.mdit import create_md_parser

parser = create_md_parser(MdParserConfig(), SphinxRenderer)
parser.options["document"] = make_document()
print(parser.render("**foo**"))
```

**expectation**
I expected content to be rendered.

**bug**

But instead here's an error message I ran into.

Note that, in Python, if the code of a `@property` raises an `AttributeError`, it gets reported as if the property is missing.

However, the problem is that the `.env` of `self.document.settings.env` in `sphinx_dev` raises an `AttributeError`.

```
Traceback (most recent call last):
  File "t.py", line 8, in <module>
    print(parser.render("**foo**"))
  File "markdown_it/main.py", line 267, in render
    return self.renderer.render(self.parse(src, env), self.options, env)
  File "myst_parser/mdit_to_docutils/base.py", line 236, in render
    self._render_finalise()
  File "myst_parser/mdit_to_docutils/base.py", line 303, in _render_finalise
    if self.sphinx_env is not None:
  File "myst_parser/mdit_to_docutils/base.py", line 127, in __getattr__
    raise AttributeError(
AttributeError: 'SphinxRenderer' object has no attribute 'sphinx_env'
```

**problem**

This is a problem for people using this library because rendering content is the point.

### Reproduce the bug

See example above

### List your environment

All latest versions allowed by MyST-Parser.

jpmckinney:
fixes #768

### DIFF
--- a/myst_parser/mdit_to_docutils/sphinx_.py
+++ b/myst_parser/mdit_to_docutils/sphinx_.py
@@ -29,8 +29,11 @@ class SphinxRenderer(DocutilsRenderer):
     """
 
     @property
-    def sphinx_env(self) -> BuildEnvironment:
-        return self.document.settings.env
+    def sphinx_env(self) -> BuildEnvironment | None:
+        try:
+            return self.document.settings.env
+        except AttributeError:
+            return None
 
     def _process_wrap_node(
         self,


jpmckinney:
Hmm, well, other methods in `SphinxRenderer` expect sphinx_env to not be done, so closing as not a complete solution.

PR #777, opened 2023-06-06
👌 IMPROVE: Substitution extension

chrisjsewell:
Allow any value type (including dict, list, datetime) and emit a `myst.substitution` warning for errors in resolving the substitution.


firasm:
Thank you, this is great!!

PR #800, opened 2023-08-06
[RFC] Rebuild link labels for Sphinx i18n support

jdknight:
When processing translations in Sphinx, after Sphinx's i8n transform replaces parts of a document with a translation, a document is processed a message at a time through the MyST-parser. For references using link labels, these may not be to the parse for the specific messages that need them and links may be left unbuilt after processing a document.

To prevent issues when using translations, cache any detected label references when processing a given document. And when individual messages are parsed after translation, rebuild any label references and append them to the message first. This will ensure references will be resolved/built as expected.

# RFC

Limited testing has been done. This appears to resolve my use case; however, I cannot say I have a complete understanding of MyST with Sphinx to know if the changes are right. If this (with maybe tweaks) works, great; if a completely different approach is needed, that's fine -- hopefully this example/workaround can provide hints to a better implementation.

# About

Noticed after converting a project from using reStructuredText to Markdown, any references built using labels would not be properly rendered when inside a part of a document that had a translation. For example, the default English document:

image

Would result in the following using stock MyST-parser with Sphinx:

image

And after applying the changes in this pull request, translated messages would work as expected:

image

The following repository can be used to demonstrate the issue:
- https://github.com/jdknight/sphinx-i8n-myst-example

### DIFF
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -948,6 +948,15 @@ class DocutilsRenderer(RendererProtocol):
         uri = cast(str, token.attrGet("href") or "")
         implicit_text: str | None = None
 
+        # cache label references for possible i8n support
+        # (see also: MystParser.parse)
+        if uri and 'label' in token.meta:
+            label = token.meta['label']
+            env = self.sphinx_env
+            docname = env.docname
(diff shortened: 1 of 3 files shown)


chrisjsewell:
Thanks I'll have a look and think about it.

Obviously it would be good to implement a test that demonstrates the current failure

PR #810, opened 2023-09-20
📚 Fix the double `used` in docs/syntax/math.md

ice-tong:
# Motivation
- The double `used` seems a typo?
https://github.com/executablebooks/MyST-Parser/blob/350c6331da4738057556d966b0bfd9d98d570352/docs/syntax/math.md?plain=1#L9

# Modification
- `docs/syntax/math.md`
  - Remove the redundant `used`.


agoose77 (approved):
Thanks for catching this!

Issue #612, opened 2022-08-15
Fix docutils 0.19 deprecation warnings


chrisjsewell:
Running the test suite (with `tox` command) now gives the following warnings:

```
tests/test_docutils.py: 9 warnings
tests/test_renderers/test_error_reporting.py: 18 warnings
tests/test_renderers/test_fixtures_docutils.py: 89 warnings
tests/test_renderers/test_fixtures_sphinx.py: 155 warnings
tests/test_renderers/test_include_directive.py: 8 warnings
tests/test_renderers/test_myst_config.py: 5 warnings
tests/test_renderers/test_myst_refs.py: 9 warnings
tests/test_sphinx/test_sphinx_builds.py: 57 warnings
  /Users/chrisjsewell/Documents/GitHub/myst-parser/myst_parser/mdit_to_docutils/base.py:55: DeprecationWarning: The frontend.OptionParser class will be replaced by a subclass of argparse.ArgumentParser in Docutils 0.21 or later.
    settings = OptionParser(components=(parser_cls,)).get_default_values()

tests/test_docutils.py: 712 warnings
tests/test_renderers/test_error_reporting.py: 1224 warnings
tests/test_renderers/test_fixtures_docutils.py: 6724 warnings
tests/test_renderers/test_fixtures_sphinx.py: 10540 warnings
tests/test_renderers/test_include_directive.py: 544 warnings
tests/test_renderers/test_myst_config.py: 760 warnings
tests/test_renderers/test_myst_refs.py: 612 warnings
tests/test_sphinx/test_sphinx_builds.py: 3876 warnings
  /Users/chrisjsewell/Documents/GitHub/myst-parser/.tox/py37-sphinx5/lib/python3.7/optparse.py:1000: DeprecationWarning: The frontend.Option class will be removed in Docutils 0.21 or later.
    option = self.option_class(*args, **kwargs)

tests/test_docutils.py::test_cli_latex
  /Users/chrisjsewell/Documents/GitHub/myst-parser/myst_parser/parsers/docutils_.py:265: FutureWarning: The default for the setting "use_latex_citations" will change to "True" in Docutils 1.0.
    _run_cli("latex", "LaTeX documents", argv)

tests/test_docutils.py::test_cli_latex
  /Users/chrisjsewell/Documents/GitHub/myst-parser/myst_parser/parsers/docutils_.py:265: FutureWarning: The default for the setting "legacy_column_widths" will change to "False" in Docutils 1.0.)
[...]

PR #842, opened 2023-12-12
🔧 Fix docutils deprecation in option parsing

agoose77:
This PR fixes the deprecation warnings raised by docutils' deprecation of `OptionParser`.

The help handling is a little ugly - we could introduce a switch on the docutils version like we do for the default settings. However, I *think* this approach will be less likely to break once the breaking change is introduced.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,6 +118,8 @@ ignore_missing_imports = true
 [tool.pytest.ini_options]
 filterwarnings = [
     "ignore:.*The default for the setting.*:FutureWarning",
+#    'ignore:.*The frontend\.OptionParser class will be replaced.*:DeprecationWarning',
+#    'ignore:.*The frontend\.Option class will be removed.*:DeprecationWarning'
 ]
 
 [tool.coverage.run]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -118,6 +118,8 @@ ignore_missing_imports = true
 [tool.pytest.ini_options]
 filterwarnings = [
     "ignore:.*The default for the setting.*:FutureWarning",
+#    'ignore:.*The frontend\.OptionParser class will be replaced.*:DeprecationWarning',
+#    'ignore:.*The frontend\.Option class will be removed.*:DeprecationWarning'
```
agoose77, line 122:
```suggestion
```

chrisjsewell:
Cheers @agoose77!

Note some suggestions on a fix were given here: https://github.com/executablebooks/MyST-Parser/issues/612#issuecomment-1265047994, maybe you want to double check that against your changes here?

agoose77:
Thanks @chrisjsewell — I actually used that issue as a starting point. As such, I'm hopeful that this should fall in line with the recommended changes where possible.

chrisjsewell (approved):
Yep I think all good cheers

PR #851, opened 2024-01-01
📚 Also add linkify to pip install command in README

n-peugnet:
See <https://github.com/executablebooks/MyST-Parser/pull/792>


chrisjsewell (approved):
Thanks!

PR #857, opened 2024-01-05
👌 IMPROVE: Allow for opening external links in new tabs (#856)

### LINKED ISSUE #820: Add support for the keys "target" and "rel" in inline_attrs
### Describe the feature you'd like to request

I would like to be able to create a link that, when clicked, opens in a new tab.

I found that [in the documentation](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-attributes), "only certain key-value attributes are supported for each syntax".

### Describe the solution you'd like

I would like `[here](example.md){target="_blank" rel="noreferer noopener"}` to work.

### Describe alternatives you've considered

My workaround is to directly use `<a href="example.html" target="_blank" rel="noreferer noopener">`

### LINKED ISSUE #856: Open external URLs in a new tab
### Describe the feature you'd like to request

Currently there is no way by using the Markdown to set a link to open on a new tab, as it is also described in https://github.com/executablebooks/MyST-Parser/issues/820. I would like to add to that, so that we create a new configuration option to add to all external URLs the setting to open to a new tab.

### Describe the solution you'd like

I propose that we add a configuration option, e.g., `external_links_new_tab`, which if set to true, all the rendered link_url will contain the attributes `target=_blank` and `rel="noreferer noopener"`.

As a bonus to the above, we will enable the `target` and `rel` keys to be configurable from the user, so that if they want to add only some of the URLs to be opened in a new tab, to be able to do so.

### Describe alternatives you've considered

_No response_

marjus45:
Add support to enable external links to open in a new tab. Expose the configuration option myst_links_external_new_tab, which if set will set all URL links to open in a new tab on the browser.

Besides that, we allow the user to set the target and rel attributes when using the extension inline_attrs (#820).

Closes #820
Closes #856

### DIFF
--- a/myst_parser/config/main.py
+++ b/myst_parser/config/main.py
@@ -233,6 +233,14 @@ class MdParserConfig:
         },
     )
 
+    links_external_new_tab: bool = dc.field(
+        default=False,
+        metadata={
+            "validator": instance_of(bool),
+            "help": "Open all external links in a new tab",
+        },
+    )
+
     url_schemes: Dict[str, Optional[UrlSchemeType]] = dc.field(
         default_factory=lambda: {
             "http": None,
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -952,8 +952,12 @@ class DocutilsRenderer(RendererProtocol):
         """
         ref_node = nodes.reference()
         self.add_line_and_source_path(ref_node, token)
+        attribute_keys = ["class", "id", "reftitle", "target", "rel"]
+        if self.md_config.links_external_new_tab:
+            token.attrs["target"] = "_blank"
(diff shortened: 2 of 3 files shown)


chrisjsewell (changes requested):
Heya thanks,

Could you add/update some test fixtures in `tests/test_renderers/test_myst_config.py`, to test both the global config and the `attrs_inline`

For example see: https://github.com/executablebooks/MyST-Parser/blob/master/tests/test_renderers/fixtures/myst-config.txt#L461

marjus45:
Thanks a lot for the pointer @chrisjsewell . I updated my commit with a couple of tests for these features. Please let me know if that will suffice!

chrisjsewell (approved):
great cheers!

PR #862, opened 2024-01-26
🔧 Add additional Ruff lints (and fix issues)

danieleades:
lints are in separate commits for ease of reviewing (not every one of these lints will necessarily be of interest)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,7 @@ exclude = [
 ]
 
 [tool.ruff]
-extend-select = ["B", "C4", "COM", "FA", "I", "ICN", "ISC", "N", "RUF", "SIM", "UP"]
+extend-select = ["B", "C4", "COM", "FA", "I", "ICN", "ISC", "N", "PIE", "RUF", "SIM", "UP"]
 extend-ignore = ["ISC001", "RUF005", "RUF012", "COM812"]
 
 [tool.mypy]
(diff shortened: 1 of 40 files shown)


pyproject.toml
```diff
@@ -102,8 +102,12 @@ exclude = [
...
 
 [tool.ruff]
-extend-select = ["B0", "C4", "I", "ICN", "ISC", "N", "RUF", "SIM", "UP"]
-extend-ignore = ["ISC001", "RUF005", "RUF012"]
+extend-select = ["B", "C4", "COM", "FA", "FURB", "I", "ICN", "ISC", "N", "PERF", "PGH", "PIE", "RUF", "SIM", "UP"]
+extend-ignore = ["ISC001", "RUF005", "RUF012", "COM812"]
+
+[tool.ruff.per-file-ignores]
+"myst_parser/parsers/docutils_.py" = ["FA"]
+"myst_parser/config/main.py" = ["FA"]
```
chrisjsewell, line 110: there is logic in these files currently utilises runtime analysis of types, so you can't have postponed evaluation

chrisjsewell (dismissed):
hope you don't mind, but I was impatient and fixed the issues causing your changes to fail CI

chrisjsewell:
Actually, I also reverted https://docs.astral.sh/ruff/rules/missing-trailing-comma/; I don't think it integrates well with the ruff formatter, in particular https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma

chrisjsewell (approved):

PR #863, opened 2024-01-26
use pre-commit to format config files

danieleades:
putting this here as a draft, because even if this is of interest discussion will be required to hammer out the specific configuration

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -1,57 +1,57 @@
 name: Bug Report
 description: Create a bug report 🐛
-labels: ['bug']
+labels: [bug]
 body:
-  - type: markdown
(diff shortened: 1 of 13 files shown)


danieleades:
closing. not a fan of the way the autoformatting handles 4-space indents in yaml

chrisjsewell:
Aha was just about to say hi, but yeh I think it's not a bad thing to have standardised formatting, but not super picky on the exact config

danieleades:
> Aha was just about to say hi, but yeh I think it's not a bad thing to have standardised formatting, but not super picky on the exact config

i'm all for standardising the formatting, but applying additional spaces on *both* sides of the hyphen when the indent is increased is pretty weird - https://github.com/executablebooks/MyST-Parser/pull/863/commits/38abd7147487412881d0fb05de2fe1a9f74b4c2d

PR #865, opened 2024-01-27
🔧 mypy- disallow 'any generics'

danieleades:
- make mypy config stricter by disallowing 'any' generics
- fix some errors by adding generics
- 'whitelist' the modules with excessive errors
- includes some changes due to autoformatting `pyproject.toml`

this whitelist should be shrunk and eventually removed over time. It should never get larger.


chrisjsewell (approved):
cheers!

PR #866, opened 2024-01-27
🔧 remove redundant mypy config

danieleades:
`no_implicit_optional` is the default value in mypy these days, so this config is redundant


chrisjsewell (approved):
sounds good

PR #904, opened 2024-04-10
📚 DOCS: Fix admonition example

72757373656c6c:
Correct admonition example to match class option definition which is a space-separated list.

Add a link to the class option definition for identifier normalization rules.

https://docutils.sourceforge.io/docs/ref/rst/directives.html#identifier-normalization


chrisjsewell (approved):
great thanks!

PR #929, opened 2024-05-04
:bug: FIX: Double encoded ampersands in query params

### LINKED ISSUE #760: Unable to link with multiple query params
### Describe the bug

**context**
I am trying to make a link containing query params such as `https://example.com?first=a&second=b`

**expectation**
I expected that the rendered link would contain all of the query params.

**bug**
But instead the `&` becomes encoded as `&amp;` and breaks the link. I thought I'd be able to escape the `&` or that it would work natively, but neither seem to be true. 

### Reproduce the bug

```
[example][EXAMPLE]

[EXAMPLE]: https://example.com?first=a&second=b "Example"
```

and

```
[example][EXAMPLE]

[EXAMPLE]: https://example.com?first=a\&second=b "Example"
```

Renders:

```html
<a class="reference external" href="https://example.com?first=a&amp;second=b" title="Example">example</a>
```

### List your environment

Using MyST-Parser: 1.0.0

sumezulike:
This closes #760.

MyST already escapes the `&` in the query params, but docutils escapes them again when rendering HTML, leading to broken query params.
```md
[Link](example.com?foo=1&bar=2)
```
becomes
```html
<p><a href="example.com?foo=1&amp;amp;bar=2">Link</a></p>
```
instead of 
```html
<p><a href="example.com?foo=1&amp;bar=2">Link</a></p>
```

These changes to the docutils parser override the `HTMLTranslator.encode` function so already encoded ampersands are not encoded again.
Since the issue lies in the interaction with docutils, I think it makes sense to fix it here.
Also, simply removing the `escapeHtml` call in MyST makes the sphinx app doctree XML files invalid because the ampersands aren't escaped.

### DIFF
--- a/myst_parser/parsers/docutils_.py
+++ b/myst_parser/parsers/docutils_.py
@@ -1,5 +1,6 @@
 """MyST Markdown parser for docutils."""
 
+import re
 from dataclasses import Field
 from typing import (
     Any,
@@ -268,6 +269,10 @@ class Parser(RstParser):
         HTMLTranslator.depart_rubric = depart_rubric_html
         HTMLTranslator.visit_container = visit_container_html
         HTMLTranslator.depart_container = depart_container_html
+        HTMLTranslator.special_chars_no_amp = {  # needed by encode_fixed
+            k: v for k, v in HTMLTranslator.special_characters.items() if k != ord("&")
+        }
+        HTMLTranslator.encode = encode_fixed
 
         self.setup_parse(inputstring, document)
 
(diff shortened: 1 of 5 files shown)


myst_parser/parsers/docutils_.py
```diff
@@ -268,6 +269,10 @@ def parse(self, inputstring: str, document: nodes.document) -> None:
         HTMLTranslator.depart_rubric = depart_rubric_html
         HTMLTranslator.visit_container = visit_container_html
         HTMLTranslator.depart_container = depart_container_html
+        HTMLTranslator.special_chars_no_amp = {  # needed by encode_fixed
+            k: v for k, v in HTMLTranslator.special_characters.items() if k != ord("&")
+        }
+        HTMLTranslator.encode = encode_fixed
```
chrisjsewell, line 275: I'm afraid this is not something that cannot be merged; it would break restructuredtext parsing.

Monkey patching is done strictly as a last resort, and only for where the patch does not affect upstream use

chrisjsewell (changes requested):

chrisjsewell:
Fixed in #1126

PR #964, opened 2024-08-19
Fix detection of YAML option block when using colon fences (fixes #963)

### LINKED ISSUE #963: Nested directives combined with colon fences do not parse YAML blocks correctly
### What version of `myst-parser` are you using?

2.0.0

### What version dependencies are you using?

docutils `0.20.1`
sphinx `7.4.7`

### What operating system are you using?

Linux

### Describe the Bug

When using this syntax:
```
::::{outer-directive}
:key: value
:::{inner-directive}
:::
::::
```

This warning message is displayed:
`WARNING: 'outer-directive': Invalid options format (bad YAML) [myst.directive_parse]`

This is because the `directives.py` file from MyST parser determines the end of the YAML block solely by looking at which lines begin with a colon.

I have a proposed fix which I will open a PR for momentarily.

### Expected Behavior

_No response_

### To Reproduce

_No response_

iso2013:
This is a quick change to fix issue #963. I think this is a reasonable solution even if the colon fence extension is not enabled, because valid YAML will never begin with `:{` (with any number of colons).

Suggestions/alternatives are welcome.

### DIFF
--- a/myst_parser/parsers/directives.py
+++ b/myst_parser/parsers/directives.py
@@ -192,7 +192,8 @@ def _parse_directive_options(
         content_lines = content.splitlines()
         yaml_lines = []
         while content_lines:
-            if not content_lines[0].lstrip().startswith(":"):
+            next_line = content_lines[0].lstrip()
+            if not next_line.startswith(":") or next_line.lstrip(":").startswith("{"):
                 break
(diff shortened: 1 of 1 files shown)


chrisjsewell:
closed in #1133 thanks

Issue #680, opened 2023-01-11
Allow at least one kind of substitution inside directive


senges:
### Context

This issue is closely related to #172, especially https://github.com/executablebooks/MyST-Parser/issues/172#issuecomment-685923790.

As @nickcrider was explaining, there is an edge case where users needs to use `{eval-rst}` but are unable to use any kind of substitution as Jinja's `{{ }}` won't be evaluated and reST substitution `.. |any| replace::` does not work either (throwing `Undefined substitution referenced: "any"` error).

To be a bit more precise, on my case, I'm using `{eval-rst}`  to handle more complex tables offered by reST (to merge cells essentially). And it's quite a known fact that tables are hard to maintain when the content gets a bit large.

Until now, with basic markdown tables I was able to use Jinja substitution :

```
---
myst:
  substitutions:
    large_text: A very long long text here that would make table super ugly...
---

| Key   | Content          |
| :---- | :--------------- |
| Story | {{ large_text }} |
```

But with reST tables, I cannot manage to make it work :

````
---
myst:
  substitutions:
    large_text: A very long long text here that would make table super ugly...
---

```{eval-rst}
+------------------------+------------+----------+----------+
| Header row, column 1   | Header 2   | Header 3 | Header 4 |
| (header rows optional) |            |          |          |
+========================+============+==========+==========+
| body row 1, column 1   | column 2   | column 3 | column 4 |
+------------------------+------------+----------+----------+
| body row 2             | {{ large_text }}                 |
+------------------------+------------+----------+----------+
```
````

````
```{eval-rst}
.. |large| replace:: A very long long text...

+------------------------+------------+----------+----------+
| Header row, column 1   | Header 2   | Header 3 | Header 4 |
| (header rows optional) |            |          |          |
+========================+============+==========+==========+
| body row 1, column 1   | column 2   | column 3 | column 4 |
+------------------------+------------+----------+----------+
| body row 2             | |large|                          |
+------------------------+------------+----------+----------+
```
````

### Proposal

I understand that this is not that simple, but that would be really great to have some kind of support for [reST substitution](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-definitions) inside `{eval-rst}` blocks.

Or any other way of doing text substitution.

I am aware that sphinx-substitution-extention supports `rst_prolog` but I cannot set these text variables inside the `conf.py` file. Each text is very page related and that would be very confusing to move it in the config file.

Maybe I'm missing something here ?
[...]

PR #966, opened 2024-08-22
Expose myst_substitions to RST

chancez:
While this doesn't fix the fact that `.. replace::` does not work within eval-rst, it does make it possible to perform substitutions within eval-rst using the same substitutions available to Jinja substitutions.

Related: #680

Honestly I have very little experience with docutils, or even MyST but I basically figured out substitutions need to be registered by calling `self.document.note_substitution_def` from reading where the `Undefined substitution referenced:` error message comes from and determining that it reads from two dictionaries that the `document.note_substitution_def` updates. I then kind of got lucky and found that `MyST` already called the necessary function for the wordcount extension, which gave me a basic example of what I needed to do. I literally just copy and pasted the code used for the word count extension but using the values from `self.md_config.substitutions` which is where `myst_substitutions` values are stored, and it "just worked".

Let me know what you think, like I said, it was really just a stab at trying to make something substitution related work for `eval-rst`. I'd still love to get `.. replace::` working but this will solve my needs pretty well I think.

### DIFF
--- a/myst_parser/mdit_to_docutils/base.py
+++ b/myst_parser/mdit_to_docutils/base.py
@@ -280,6 +280,17 @@ class DocutilsRenderer(RendererProtocol):
                     substitution_node, f"wordcount-{key}"
                 )
 
+        for key, value in self.md_config.substitutions.items():
+            if value is None:
+                continue
+
+            substitution_node = nodes.substitution_definition(
+                str(value), nodes.Text(str(value))
+            )
+            substitution_node.source = self.document["source"]
+            substitution_node["names"].append(key)
+            self.document.note_substitution_def(substitution_node, key)
+
     def nested_render_text(
         self,
         text: str,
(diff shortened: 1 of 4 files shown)


MacqGit:
Do what I expected it would do.

In my case, I have lots of "reusable" documents where I use 'substitutions' to eventually generate the right output when product names etc... get eventually defined. However I also intensively use "conditional compiling" (in RST) to get parts of docs included/excluded from the final output.
I love using Myst-Parser but the lack of substitutions within RST refrained me to fully transition them to MD.

Thanks a lot for the time spend and the solution.

Bizordec:
@chrisjsewell please review this PR 🙏 

I think it would help to implement support for frontmatter substitutions in https://github.com/adamtheturtle/sphinx-substitution-extensions (https://github.com/adamtheturtle/sphinx-substitution-extensions/issues/1554)

adamtheturtle:
I rebased this implementation onto current `master`, preserved @chancez’s commit authorship, added coverage for the `None` value branch, and opened #1194 as a successor because this account cannot push to the contributor fork despite `maintainer_can_modify` being enabled.

chancez:
I'll close my PR in favor of @adamtheturtle's rebased implementation.

PR #987, opened 2024-10-23
📚 remove redundant paragraph

Snoopy1866:
I think this paragraph is redundant here.

image


chrisjsewell (approved):
Cheers

Snoopy1866:
It appears that some continuous integration processes have failed, but reviewing the console logs hasn't provided any clarity on the issue. In theory, simply modifying the documentation should not cause such failures, I think.

Is there any other situation which need to be consider? 🤔

chrisjsewell:
all sorted cheers

Issue #1017, opened 2025-01-21
Support using `beartype` and MyST-Parser for creating a custom role which requires an `Inliner`


adamtheturtle:
### Goals

* Create a custom role which extends `docutils.parsers.rst.roles.code_role`.
* Pass static type checking (`mypy`)
* Pass runtime type checking (`beartype`)

Below is a minimized example of my work towards supporting MyST for `substitution-code` within https://github.com/adamtheturtle/sphinx-substitution-extensions.

### Requirements

* `beartype==0.19.0`
* `myst-parser==4.0.0`
* `types-docutils==0.21.0.20241128`

### Files

Layout as described in
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions.

```
<project directory>/
├── conf.py
└── sphinxext/
    └── extname.py
```

```python
# conf.py

import sys
from pathlib import Path

# Path configured as per
# https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions
sys.path.append(str(Path('sphinxext').resolve()))

extensions = [
    "myst_parser",
    "extname",
]
```

```python
# sphinxext/extname.py

from typing import Any

from beartype import beartype
from docutils.nodes import Node, system_message
from docutils.parsers.rst.roles import code_role
from docutils.parsers.rst.states import Inliner


# Typed version of the example from https://docutils.sourceforge.io/docs/howto/rst-roles.html#define-the-role-function
@beartype
def role_fn(
    name: str,
    rawtext: str,
    text: str,
    lineno: int,
    # This is an `Inliner`, so it can be passed through directly to `code_role`.
    inliner: Inliner,
    options: dict[Any, Any] = {},
    content: list[str] = [],
) -> tuple[list[Node], list[system_message]]:
    print("Custom logic")
    # Type stubs for `code_role` are at
    # https://github.com/python/typeshed/blob/57d7c4334b64856fda6f6e8f992b101ddafe2f57/stubs/docutils/docutils/parsers/rst/roles.pyi#L103
    #
    # Importantly, they require that `inliner` is an `Inliner`.
    return code_role(   
        role=name,
        rawtext=rawtext,
        text=text,
        lineno=lineno,
        inliner=inliner,
        options=options,
        content=content,
    )

@beartype
def setup(app):
    app.add_role(name="my-role", role=role_fn)
```

### Command

```shell
sphinx-build -M html . build/
```

### Error

```
[...]

PR #1018, opened 2025-01-21
Enable using MockInliner in place of `Inliner` for runtime type checking

adamtheturtle:
Fixes https://github.com/executablebooks/MyST-Parser/issues/1017.

### DIFF
--- a/myst_parser/mocking.py
+++ b/myst_parser/mocking.py
@@ -34,6 +34,16 @@ class MockInliner:
     This is parsed to role functions.
     """
 
+    # Override the ``__class__`` attribute to enable runtime type checking.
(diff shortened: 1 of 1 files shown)


myst_parser/mocking.py
```diff
@@ -34,6 +34,16 @@ class MockInliner:
     This is parsed to role functions.
     """
 
+    # Override the ``__class__`` attribute to enable runtime type checking.
+    # See https://beartype.readthedocs.io/en/latest/faq/#mock-types.
+    @property
+    def __class__(self) -> type:
+        return Inliner
+
+    @__class__.setter
```
adamtheturtle, line 43: This is not (yet) in the Beartype documentation, but the explanation for it is in https://github.com/beartype/beartype/pull/485.

Issue #968, opened 2024-08-26
Support stable heading id for non english languages


Alexey-NM:
### Describe the feature you'd like to request

I need ability to make refrences to headers from external documentation (some_chapter.html#zagalovok).
Unfortunately If I have a title in a non english language if will have numeric id which can be changed in any time. 
It is difficult to automate and it is unstable because that number can change at any time.
```html
<section id="id2">
```

### Describe the solution you'd like

I want to have the ablility to define the id  generation function  in my conf.py to a function that accepts a string and returns a stable id.

The function will generate stable id by  a node title.
```html
<section id="zagalovok">
```

I have maken some test in the function [generate_heading_target](https://github.com/executablebooks/MyST-Parser/blob/1038800fc411296369604a03756bec6945c6d4d8/myst_parser/mdit_to_docutils/base.py#L776)

```py
name = nodes.fully_normalize_name(implicit_text)
name = conf.make_id_by_any_lang(name)
```
It seems It is working.

Will it be accepted If I make merge request?

Alexey-NM:
Alternative decision is to add ability configuring custom markdown plugins.
A plugin can add id to token similar to [attrs_block extension](https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#block-attributes)

Alexey-NM:
>what exactly does make_id_by_any_lang do, how is it different to

It generates an [english name from other language](https://pypi.org/project/transliterate/). It allows to use static id.  

```py
def make_id_by_any_lang(str:str)->str:
    # see https://pypi.org/project/transliterate/
    tr_str = translit(str,"ru",reversed=True)
    #This function is a modified version of ``docutils.nodes.make_id()`` of
    #docutils-0.16
    # It is test code so probably it is not nessasary
    result = _make_id(tr_str).lower()
    return result
```

Alexey-NM:
We have implemented [pull request](https://github.com/executablebooks/MyST-Parser/pull/1019) which allows to do it

PR #1019, opened 2025-01-27
Add ability to generate stable heading id for non english languages by setting myst_fully_normalize_name_slug_func

Alexey-NM:
They need stable heading id to make refrences to headers from external documentation (some_chapter.html#zagalovok).
Unfortunately If  there are  titles in a non english language it will have numeric id which can be changed in any time.
It is difficult to automate and it is unstable because such number can be changed at any time.

This merge request allows to generate stable ids by setting myst_fully_normalize_name_slug_func in config.py
```python
myst_fully_normalize_name_slug_func = 'utils.fully_normalize_name_by_any_lang'
```

```
def fully_normalize_name_by_any_lang(str:str)->str:
    normalize_name = ' '.join(str.lower().split())
    tr_str = translit(normalize_name,"ru",reversed=True)
    result = _make_id(tr_str)
    return result
```
It generates an [english name from other language](https://pypi.org/project/transliterate/). It allows to use static id.

See also: https://github.com/executablebooks/MyST-Parser/issues/968#issuecomment-2312571907

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,43 @@
+name: Copilot Setup Steps
+
+# Automatically run the setup steps when they are changed to allow for easy validation, and
+# allow manual testing through the repository's "Actions" tab
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+jobs:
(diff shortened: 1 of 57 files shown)


chrisjsewell:
Closing in favour of #1158, which addresses the underlying problem (#968) differently: the GitHub-style slug — including for non-English headings, e.g. `#заголовок` — is now emitted as a real, stable HTML anchor, and `myst_heading_slug_func` gains named presets, so no second callable-based config option is needed (we are also moving configuration towards plain declarative values). Thanks for the contribution and for highlighting the non-English-ids gap — it directly shaped that work.

---

PR #1036, opened 2025-05-09
📚 Fix minor typo in `cross-referencing.md`

sebras:
I think I've followed the pattern of other contributions. Please let me know if I've missed something.


chrisjsewell (commented):
cheers! (better late than never 😅 )

chrisjsewell (approved):

PR #1114, opened 2026-03-20
🐛 FIX: Use docname instead of source path in warning locations

Bizordec:
Changed warning location generation to use `document.settings.env.docname` instead of `document["source"]`. This ensures warning locations display clean file names (e.g., `index.md`) rather than full paths with potential `.rst` suffixes (e.g., `index.md.rst`).


chrisjsewell (approved):
Looks good thanks!

PR #1158, opened 2026-07-12
✨ NEW: stable heading anchors: shared slug module, presets, HTML ids, explicit-id priority

### LINKED ISSUE #527: Auto-generated header links ignore mis-anchor headers that start with a number
### Describe the bug

When section headers start with a number, they are labeled in a confusing way by the auto-header anchor generator. They simply have the structure `id<num>`.

For example, a document like this:

```md
## v2.0

## 2.0
```

Will generate anchors like:

```
<a class="headerlink" href="#v2-0" title="Permalink to this headline">
<a class="headerlink" href="#id1" title="Permalink to this headline">
```

I think users would instead expect the ID of the second element to be `2-0`

Since we describe the behavior as based on the GitHub sluggify, I also quickly checked and it looks like they do support IDs that start with numbers (that's why I'm labeling this as a bug since I think it breaks from expected behavior based on that criteria).

I think the relevant code for this might be here:

https://github.com/executablebooks/mdit-py-plugins/blob/master/mdit_py_plugins/anchors/index.py#L119-L120

originally reported in: https://github.com/jupyterhub/kubespawner/pull/575

### Reproduce the bug

1. Set `myst_heading_anchors = 2`
2. Create a page like 

	  ```md
      # Test
	  ## v2.0
	  
	  ## 2.0
	  ```
3. Render it and note the heading IDs

### List your environment

_No response_

### LINKED ISSUE #828: Add support for GitLab slugger
### Describe the feature you'd like to request

The GitLab slugger’s source is [here](https://gitlab.com/gitlab-org/gitlab/-/tree/master/lib/gitlab/slug), it would be great if MyST provided that function to use as a drop-in replacement for the GitHub one.

This would be helpful to me, since I want my links to work both in GitLab Markdown files and in MyST.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### LINKED ISSUE #878: third occurence of heading with the same title cannot be referenced, `[myst.xref_missing]`
### What version of `myst-parser` are you using?

2.0.0

### What version dependencies are you using?

sphinx : 6.2.1
docutils 0.19

### What operating system are you using?

Linux

### Describe the Bug

In the docs (especially here https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#implicit-targets) it says that the slugs for headings matches the github slugger.
As such, when headings have the same name, the second occurence gets a suffix of `-1`, and the third should get `-2` and so on.

It looks like this works for myst parser but only for two headings. As soon as there are 3 headings with the same name (can happen with subsections that follow a pattern between sections), any cross ref fails

Capture d’écran du 2024-02-26 23-40-38

See how I get the warning `'myst' reference target not found: 'heading-2' [myst.xref_missing]` even though the slug does exist in the html.

### Expected Behavior

`[Link to the heading 3](#heading-2)` should link to the third occurence of the heading titled "heading" and not issue a warning

### To Reproduce

This bug is very to reproduce since it happens on the live preview
https://myst-parser.readthedocs.io/en/latest/live-preview.html#

```md
# Heading

Hallo world!

[Link to the heading](#heading)

[Link to the heading 2](#heading-1)

[Link to the heading 3](#heading-2)

# Heading

The second heading

# Heading

the third heading
```

I kept the default configuration, i.e.
```yaml
myst_enable_extensions:
- colon_fence
- deflist
- dollarmath
myst_heading_anchors: 2
myst_highlight_code_blocks: true
```

### LINKED ISSUE #885: WARNING: 'myst' cross-reference target not found: 'level-4-header-title' [myst.xref_missing]
### What version of `myst-parser` are you using?

2.0.0

### What version dependencies are you using?

Name: Sphinx
Version: 7.2.6

Name: docutils
Version: 0.20.1

### What operating system are you using?

Mac

### Describe the Bug

When internal linking to a fourth level header (\#\#\#\# Fourth Level Header) using the syntax `[link text](#fourth-level-header)`, upon doc compilation myst throws this error the first time during compilation:  `WARNING: 'myst' cross-reference target not found: 'fourth-level-header' [myst.xref_missing]`.  However the link works when inspecting and testing the compiled docs.   

### Expected Behavior

An error will not be thrown. 

### To Reproduce

Try to compile this file below with a command such as `sphinx-build -b html src build`:

```
# First

[link to first](#first)

## Second

[link to second](#second)

### Third

[link to third](#third)

#### Fourth

[link to fourth](#fourth)
```

results in error: 

```
WARNING: 'myst' cross-reference target not found: 'fourth' [myst.xref_missing]
```

### LINKED ISSUE #908: replacement of "." in headings for slugs
### What version of `myst-parser` are you using?

2.0.0

### What version dependencies are you using?

Sphinx 7.2.6

### What operating system are you using?

Linux

### Describe the Bug

Anchors are generated differently in Myst parser and GitHub markdown parsers. Specifically,
`# Ubuntu 20.04` geta the anchor `ubuntu-2004` on GitHub, but `ubuntu-20-04` by Myst parser

### Expected Behavior

Same anchor as GitHub.

### To Reproduce

See above

### LINKED ISSUE #968: Support stable heading id for non english languages

### LINKED ISSUE #985: Literal . (dot) in header yields local id not found [myst.xref_missing]
### What version of `myst-parser` are you using?

2.0.0

### What version dependencies are you using?

Sphinx v7.3.7

### What operating system are you using?

Mac

### Describe the Bug

(ATTN: Shortened examples)

File `container_intro.md` contains:

```
# lxc.env for environment variables
```

I have another document `container_from_scratch.md` where I want to reference this header in a link like
```
Please see the [container intro about environment](./container_intro.md#lxc-env-for-environment-variables)
```

This produces the following error:
```
[...]
writing output... [100%] index
docs/container/container_from_scratch.md:304: WARNING: local id not found in doc 'docs/container/container_intro': 'lxc-env-for-environment-variables' [myst.xref_missing]
generating indices... genindex done
[...]
```

Other combinations also do NOT work, e.g. adding an explicit `(lxc-env)=` target
```
[](container_intro.md#lxc-env)
```

**:warning: The only way to resolve this WARNING is to REMOVE the dot from the header, e.g.**
```
### lxcenv for environment variables
```
and changing the link to `container_intro.md#lxc-env-for-environment-variables`.

**:warning: NOTE: The links work in the end, it is just a distracting, red WARNING**

### Expected Behavior

No WARNING is issued, parsing works.

### To Reproduce

Create 2 documents like above and compile

### LINKED ISSUE #1073: Implicit header slugs are used (e.g. in sidebar) instead of explicit ones
### What version of `myst-parser` are you using?

4.0.1

### What version dependencies are you using?

sphinx                        8.2.3
docutils                      0.21.2

### What operating system are you using?

Linux

### Describe the Bug

Even when setting `myst_heading_anchors = 0`, the headings get auto-generated (=unstable) HTML `id`s.

This would be OKish if those could only be figured out by inspecting the HTML, but they seem to be used as the primary ID in the HTML (e.g. in sphinx-book-theme’s sidebar).

### Expected Behavior

When an explicit header target is set using the `(foo)=` syntax, that one should be the canonical one used by the sidebar and in `objects.inv` (I think the latter is probably already the case)

When implementing this, you should document if the first or the last explicit target is treated as the canonical one.

### To Reproduce

```markdown

(foo)=
# Bar
```

the above should render to `<h1 id="foo">Bar</h1>` and nothing should have the ID “bar”.

chrisjsewell:
Fixes #878, fixes #985, fixes #908, fixes #527, fixes #968, fixes #1073, fixes #885, fixes #828.

Overhauls heading-anchor slugs so `myst_heading_anchors` behaves as documented ("GitHub-style slugs"), with one shared implementation, anchors that actually exist in published HTML, no false warnings for working links, and named slugger presets. Supersedes #1019 (see below).

## What changed

**One slug implementation** (`myst_parser/slugs.py`, new): pure/stdlib-only functions (`github_slugify`, `gitlab_slugify`, `unique_slug`, `SLUG_PRESETS`) shared by the renderer and the `myst-anchors` CLI — previously two parallel copies that drifted. The unit-test corpus is a language-agnostic data file (`tests/fixtures/slugs.json`), so alternative implementations can consume it verbatim.

**#878 — duplicate suffixing fixed**: four identical `# Heading`s now produce `heading, heading-1, heading-2, heading-3` (the documented GitHub behaviour), not `heading, heading-1, heading-1-2, heading-1-2-3`. A renderer/CLI cross-check test over a 13-heading corpus guards against future drift.

**#908/#527/#968 — slugs become real HTML anchors**: the GitHub-style slug (`ubuntu-2004`, `заголовок`, `20`) is now *additionally* emitted as a secondary id, so the documented anchor exists in published output. Crucially this is done by a dedicated transform (`AddSlugIds`, priority 700) that runs after **all** other id assignment, so: the docutils id stays the section's primary id (tocs, permalinks, `objects.inv`, searchindex are byte-identical); a slug can never claim an id that docutils or the user would assign to another element; and sphinx's `SortIds` cannot rotate the slug into primary position for auto-`idN` headings. Escape hatch: `myst_heading_anchors_html_ids = False` restores lookup-only behaviour. Slugs from custom `slug_func`s containing whitespace (invalid in HTML ids) are not emitted.

**#1073 — explicit targets win**: a new `PrioritiseExplicitIds` transform (priority 261, directly after docutils' `PropagateTargets`) makes an explicit `(name)=` target propagated onto a section its *primary* id, so sidebars, permalinks and `objects.inv` use the stable explicit id. The implicit heading id is kept as a secondary anchor, so previously published fragments keep working. `{#custom}` attrs ids already behaved correctly and are unchanged.

**#985/#885 — no more warnings for working links**:
- same-document `[](#target)`: implicit docutils names/ids (e.g. headings beyond the `heading_anchors` depth) now resolve instead of warning — in docutils mode directly; under sphinx only as a **last resort** after project-wide and intersphinx resolution both fail, so links that previously resolved to another document's explicit target keep doing so (regression-tested).
- cross-document `[](./doc.md#id)`: ids that demonstrably exist in the target document (section ids, std-domain label ids) resolve silently; a std label referenced by *name* now resolves to its actual anchor id (previously: warning + broken fragment).

**#828 — named presets**: `myst_heading_slug_func = "github"` (the default algorithm) or `"gitlab"` are now the canonical, declarative configuration form; dotted import paths remain as a legacy escape hatch, and unknown dotless strings get a clear error naming the presets (previously an uncaught `ValueError`). The `myst-anchors` CLI gains `--slug-func github|gitlab`. Note: the gitlab preset mirrors `Gitlab::Utils::Markdown#string_to_anchor` (strip, lowercase, remove non-word, squeeze hyphens, `anchor-` prefix for digit-only) — verified against the algorithm source statically; a sanity check against a live GitLab render before release would be prudent.

## Compatibility

The only behaviour changes for previously-building projects are: (1) third-and-later duplicate slugs (`x-1-2` → `x-2`; the old output contradicted the documented contract — bugfix); (2) additional secondary anchors appearing (purely additive); (3) the explicit-target primary-id change above (old fragments keep working via the retained implicit id; called out in the docs with a `versionchanged`); (4) links that previously **warned** now resolving when the anchor exists. Everything else was verified byte-identical old-vs-new, including: all published fragments and resolved hrefs, toc/searchindex anchors (incl. non-ASCII/digit headings under sphinx `SortIds`), `objects.inv` for projects without explicit targets, and cross-document resolution precedence. epub and latex builds verified with non-ASCII/digit-leading secondary ids.

The documented `#`-destination resolution order in `docs/syntax/cross-referencing.md` has been corrected to match actual behaviour (the old text claimed a local-implicit step that never existed) with the new last-resort step appended.

## Review

The branch went through a full adversarial + behavior-regression review (old-vs-new comparison on real docutils and sphinx builds) plus a mechanical correctness/testing/performance pass; the third commit fixes everything found, each with a regression test — notably the id-stealing, `SortIds` promotion, and resolution-precedence issues that the initial implementation had.

Suite: 1156 passed locally; own docs build clean.

## Suggested follow-ups (out of scope)

- Upstream `mdit-py-plugins`: drop `.strip()` in the anchors-plugin slugify to converge fully with the renderer.
- PR #1019 should be closed in favour of this (its goal — stable ids for non-English headings — is met by the slug anchors + existing `heading_slug_func`, without a second callable-based config option).

### DIFF
--- a/docs/syntax/cross-referencing.md
+++ b/docs/syntax/cross-referencing.md
@@ -69,7 +69,7 @@ Headings within documents can also be assigned an implicit target,
 by setting the `myst_heading_anchors` configuration option.
 This is should be set to an integer, between 1 and 6, indicating the depth of headings to assign targets to.
 
-The anchor "slugs" are created according to the [GitHub implementation](https://github.com/Flet/github-slugger): heading titles are lower cased, punctuation is removed, spaces are replaced with `-`, and uniqueness is enforced by suffix enumeration.
+By default the anchor "slugs" are created according to the [GitHub implementation](https://github.com/Flet/github-slugger): heading titles are lower cased, punctuation is removed, spaces are replaced with `-`, and uniqueness is enforced by suffix enumeration (`x`, `x-1`, `x-2`, ...). The slug function is configurable via `myst_heading_slug_func` (e.g. the `"gitlab"` preset); see [auto-generated header anchors](#syntax/header-anchors).
 
 For example, using `myst_heading_anchors = 2`:
 
--- a/docs/syntax/cross-referencing.md
+++ b/docs/syntax/cross-referencing.md
@@ -17,6 +17,12 @@ There are three primary ways to create targets:
 2. Annotating a syntax block/inline/span with an `{#id}` attribute (using the [attrs_block](#syntax/attributes/block) and [attrs_inline](#syntax/attributes/inline) extensions)
 3. Adding a `name` option to a directive
 
+```{versionchanged} 5.2.0
+An explicit target placed before a section heading now becomes the section's
+primary HTML id, used by tocs, permalinks and `objects.inv`; the implicit
+heading id is kept as a secondary anchor, so existing fragment links keep working.
+```
+
 ::::{myst-example}
 
 (heading-target)=
@@ -157,10 +163,15 @@ By default, MyST will resolve link destinations according to the following rules
 
    {style=lower-roman}
    1. First, explicit targets in the same file are searched for, if not found
-   2. Then, implicit targets in the same file are searched for, if not found
+   2. Then, implicit [heading anchors](#syntax/header-anchors) in the same file are searched for, if not found
    3. Then, explicit targets across the whole project are searched for, if not found
    4. Then, intersphinx references are searched for, if not found
-   5. A warning is emitted and the destination is left as an external link.
+   5. Then, any other anchor that exists in the same file is used (e.g. a heading beyond the `myst_heading_anchors` depth), if not found
+   6. A warning is emitted and the destination is left as an external link.
+
+   ```{versionchanged} 5.2.0
+   Step v was added: such links previously emitted a warning.
+   ```
 
 :::{note}
 Local file path resolution and cross-project references are not available in [single page builds](#myst-docutils)
--- a/docs/syntax/optional.md
+++ b/docs/syntax/optional.md
@@ -528,27 +528,43 @@ The paths to other files should be relative to the current file:
 
 ### Anchor slug structure
 
-The anchor "slugs" created aim to follow the [GitHub implementation](https://github.com/Flet/github-slugger):
+By default the anchor "slugs" created aim to follow the [GitHub implementation](https://github.com/Flet/github-slugger):
 
 - lower-case text
 - remove punctuation
 - replace spaces with `-`
-- enforce uniqueness *via* suffix enumeration `-1`
+- enforce uniqueness *via* suffix enumeration: duplicates of `x` become `x`, `x-1`, `x-2`, ... (the base slug is never mutated)
 
-To change the slug generation function, set `myst_heading_slug_func` in your `conf.py` to a function that accepts a string and returns a string.
+The slug generation function is selected with the `myst_heading_slug_func` option in your `conf.py`.
+The recommended form is a named preset:
 
-:::{versionadded} 0.19.0
-`myst_heading_slug_func` can now also be set to a string,
-which will be interpreted as an import path to a function,
-e.g. `myst_heading_slug_func = "mypackage.mymodule.slugify"`.
+```python
+myst_heading_slug_func = "github"  # the default; also available: "gitlab"
+```
+
+The `"gitlab"` preset follows [GitLab's algorithm](https://docs.gitlab.com/ee/user/markdown.html#heading-ids-and-links) (which, unlike GitHub, strips surrounding whitespace, squeezes repeated hyphens and prefixes digit-only slugs with `anchor-`).
+
+:::{versionchanged} 5.2.0
+`myst_heading_slug_func` now accepts the preset names `"github"` (default) and `"gitlab"`.
+
+As a legacy form, it can still be set to an import path string
+(e.g. `myst_heading_slug_func = "mypackage.mymodule.slugify"`, added in 0.19.0)
+or, in `conf.py`, directly to a callable that accepts a string and returns a string.
+:::
+
+:::{note}
+Anchor slugs are additionally emitted as (secondary) HTML `id` attributes on the rendered headings, so that `#slug` fragments resolve in the published HTML.
+The primary docutils id is left first and unchanged, so previously published fragments keep working.
+Set `myst_heading_anchors_html_ids = False` to disable this and emit only the docutils id.
 :::
 
 ### Inspect the links that will be created
 
-You can inspect the links that will be created using the command-line tool:
+You can inspect the links that will be created using the command-line tool
+(use `--slug-func` to preview a preset, matching `myst_heading_slug_func`):
 
 ```console
-$ myst-anchors -l 2 docs/syntax/optional.md
+$ myst-anchors -l 2 --slug-func github docs/syntax/optional.md
 <h1 id="optional-myst-syntaxes"></h1>
 <h2 id="admonition-directives"></h2>
 <h2 id="auto-generated-header-anchors"></h2>
--- a/myst_parser/cli.py
+++ b/myst_parser/cli.py
@@ -7,6 +7,7 @@ from mdit_py_plugins.anchors import anchors_plugin
 
 from myst_parser.config.main import MdParserConfig
 from myst_parser.parsers.mdit import create_md_parser
+from myst_parser.slugs import SLUG_PRESETS
 
 
 def print_anchors(args=None):
@@ -29,9 +30,17 @@ def print_anchors(args=None):
     arg_parser.add_argument(
         "-l", "--level", type=int, default=2, help="Maximum heading level."
     )
+    arg_parser.add_argument(
+        "--slug-func",
+        choices=sorted(SLUG_PRESETS),
+        default="github",
+        help="Slug preset for anchor ids (matches myst_heading_slug_func presets).",
+    )
     args = arg_parser.parse_args(args)
     parser = create_md_parser(MdParserConfig(), RendererHTML)
-    parser.use(anchors_plugin, max_level=args.level)
+    parser.use(
+        anchors_plugin, max_level=args.level, slug_func=SLUG_PRESETS[args.slug_func]
+    )
 
     def _filter_plugin(state: StateCore) -> None:
         state.tokens = [
--- a/myst_parser/config/main.py
+++ b/myst_parser/config/main.py
@@ -152,7 +152,7 @@ def check_heading_slug_func(
                 module_path, function_name = value.rsplit(".", 1)
                 mod = import_module(module_path)
                 value = getattr(mod, function_name)
-            except ImportError as exc:
+            except (ImportError, AttributeError, ValueError) as exc:
                 raise TypeError(
                     f"'{field.name}' could not be loaded from string: {value!r}"
                 ) from exc
--- a/myst_parser/config/main.py
+++ b/myst_parser/config/main.py
@@ -138,19 +138,30 @@ def check_inventories(_: "MdParserConfig", field: dc.Field, value: Any) -> None:
 def check_heading_slug_func(
     inst: "MdParserConfig", field: dc.Field, value: Any
 ) -> None:
-    """Check that the heading_slug_func is a callable."""
+    """Check that the heading_slug_func is a preset name or a callable."""
     if value is None:
         return
     if isinstance(value, str):
-        # attempt to load the function as a python import
-        try:
-            module_path, function_name = value.rsplit(".", 1)
-            mod = import_module(module_path)
-            value = getattr(mod, function_name)
-        except ImportError as exc:
+        from myst_parser.slugs import SLUG_PRESETS
+
+        if value in SLUG_PRESETS:
+            value = SLUG_PRESETS[value]
+        elif "." in value:
+            # attempt to load the function as a python import (legacy form)
+            try:
+                module_path, function_name = value.rsplit(".", 1)
+                mod = import_module(module_path)
+                value = getattr(mod, function_name)
+            except ImportError as exc:
+                raise TypeError(
+                    f"'{field.name}' could not be loaded from string: {value!r}"
+                ) from exc
+        else:
             raise TypeError(
-                f"'{field.name}' could not be loaded from string: {value!r}"
-            ) from exc
+                f"'{field.name}' is not a known preset "
+                f"({', '.join(sorted(SLUG_PRESETS))}) "
+                f"or a python import path: {value!r}"
+            )
         setattr(inst, field.name, value)
     if not callable(value):
         raise TypeError(f"'{field.name}' is not callable: {value!r}")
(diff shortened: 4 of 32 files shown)


chrisjsewell (commented):
## Review: stable heading anchors

I reviewed the full diff and exercised it locally (Sphinx 9.0.4 / docutils 0.22.4): the whole suite passes (**1156 passed, 11 skipped**), and all 21 CI checks are green across the py3.11–3.14 × Sphinx 8/9 × {ubuntu, windows} matrix. This is a well-architected, carefully-tested change and I'd be comfortable merging it after considering the notes below (none are blocking).

### Overview
Consolidates heading-anchor slugging into one stdlib-only module (`slugs.py`), fixes the duplicate-suffix contract (`x, x-1, x-2`), emits slugs as *secondary* HTML ids via a late `AddSlugIds` transform, promotes explicit `(name)=` targets to the primary id via `PrioritiseExplicitIds`, and stops warning on `#`-links that actually resolve. Closes 8 issues.

### What's strong
- **Single source of truth + language-agnostic corpus.** `slugs.py` being pure/`re`-only with `tests/fixtures/slugs.json` (guarded by `test_slugs_module_has_no_third_party_deps`) is a genuinely good design for cross-implementation parity, and the renderer/CLI cross-check (`test_anchor_slugs_match_renderer`) prevents the exact drift this replaces.
- **Transform-ordering discipline.** Doing id emission at priority 700 (after *all* id assignment) rather than at parse time is the right call — the `test_slug_id_cannot_steal_later_ids` and `test_slug_id_stays_secondary_under_sortids` regressions capture the two subtle failure modes precisely. I verified the id-stealing guard (`slug not in self.document.ids`) behaves correctly against both a later section's docutils id and a user's explicit target.
- **Resolution precedence.** `test_project_resolution_beats_local_implicit` locks down the important invariant that the new local fallback is genuinely last-resort under Sphinx (recorded as `reflocalid`, used only after project-wide + intersphinx fail). The docs resolution-order list now matches the code.

### Correctness — verified, no defects found
- Slug algorithms match references: github duplicate suffixing, and gitlab (`a---b`→`a-b`, `2.0`→`anchor-20`, `---`→`-`, digit-only prefixing) all check out against `Gitlab::Utils::Markdown#string_to_anchor`.
- `unique_slug` loop is bounded by `len(existing)`; `compute_unique_slug`'s `Iterable`→`Container` type fix is correct (it uses `in`).
- I specifically probed the adversarial `PrioritiseExplicitIds` case — an explicit target whose id *starts with* `id` (e.g. `(idea)=`), which `SortIds` would rotate to the back — and it resolves correctly (`ids=[idea, my-heading]`). See the one robustness note below.

### One robustness note (worth addressing)
- **`PrioritiseExplicitIds` shares `default_priority = 261` with Sphinx's `SortIds`.** It currently works only because docutils breaks equal-priority ties by insertion order and MyST's parser transforms are appended *after* the Sphinx/reader transforms, so `PrioritiseExplicitIds` happens to apply last and gets the final say. That's a latent fragility: a future Sphinx change to `SortIds`' priority or registration point could silently flip the order and let `SortIds` rotate an `id`-prefixed explicit id back off the front. Bumping to **`default_priority = 262`** (strictly after both `PropagateTargets` 260 and `SortIds` 261) makes the intended ordering explicit and order-independent, with no effect on `AddSlugIds` (700) / `ResolveAnchorIds` (879). Cheap insurance.

### Performance
- `AddSlugIds` walks the whole doctree once (fine); it iterates *all* nodes incl. text nodes — filtering to `nodes.Element`/`nodes.section` (plus the document node) would trim it, but the impact is negligible.
- The cross-doc fallback does linear scans per unresolved ref: `any(sect_id == ref_id …)` over the target doc's headings, then `_std_label_id_in_doc` over **all** `labels`+`anonlabels`. Only hit on the miss path, so acceptable, but a project with many broken-ish `[](./doc.md#id)` links pays O(refs × labels). If it ever shows up, a per-docname label index built once would remove it.

### Durability / compatibility
- The headline behavior change is #1073: explicit target → primary id, which **changes `objects.inv` anchors and permalinks** for projects that put `(name)=` before headings. It's the intended fix, the implicit id is retained as a secondary anchor, and it's documented with a `versionchanged` — well handled, but it's the one thing downstream projects deep-linking via `objects.inv` could notice.
- `check_heading_slug_func` legacy import path still only catches `ImportError` — `myst_heading_slug_func="os.nope"` raises a raw `AttributeError`, and `".foo"` raises `ValueError`, both bypassing the friendly `TypeError`. Pre-existing, but since the PR rewrites this block, widening to `except (ImportError, AttributeError, ValueError)` would be a tidy freebie.

### Testing — strong, a couple of small gaps
- Coverage is thorough. Minor additions worth considering: the `myst_heading_anchors_html_ids=False` escape hatch is only tested under docutils (not Sphinx); the whitespace-slug skip in `AddSlugIds` is only asserted implicitly via the `myst-config` fixture; `_std_label_id_in_doc`'s name-vs-id precedence is untested (very edge).

### Minor / nits
- `slugs.py` writes the CJK range as literal ideographs `[一-鿿]` where the code it replaces used `[一-鿿]`. The escape form is more greppable and encoding-robust in review/editors — consider keeping it. (Separately, Python's Unicode `\w` already matches CJK, so the explicit range may be redundant — but that's a pre-existing question, not for this PR.)
- `run()`'s `reflocalid` fallback builds `nodes.reference()` without `internal=True`, unlike `make_refnode`; the `refid` still resolves same-document, but setting it would keep the `internal` class consistent with other resolved refs.
- `default_slugify` is soft-deprecated (docstring only). Fine, but if you want tooling to flag it, a `DeprecationWarning` would help. Trivial: the docstring says `.. deprecated:: 5.2` while the docs use `5.2.0`.

### Security
No concerns. Slug regexes are linear (no ReDoS), `unique_slug` is bounded, and the only `import_module` path is config-driven (trusted) and now bypassed entirely by the presets.

Nice work — the adversarial/regression pass clearly paid off, and the third commit's fixes address the genuinely tricky id-lifecycle issues.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

---

chrisjsewell (commented):
## Follow-up: review notes addressed ✅

`f7e91de` (+ test fixup `1b920da`) addresses essentially every actionable point from my earlier review. I verified each against the diff and re-ran the suite (**1158 passed**, 11 skipped — the two added tests, no regressions).

| Finding | Fix | |
|---|---|---|
| **`PrioritiseExplicitIds` shared priority 261 with Sphinx `SortIds`** (main robustness note) | `default_priority = 262`, strictly after both — no longer insertion-order dependent | ✅ |
| `AddSlugIds` iterated all nodes incl. text (perf) | `findall(...)(nodes.Element)`, redundant `isinstance` dropped | ✅ |
| `check_heading_slug_func` only caught `ImportError` (durability) | `except (ImportError, AttributeError, ValueError)` — `"os.nope"` / `".foo"` now give the friendly `TypeError` | ✅ |
| `reflocalid` fallback ref missing `internal=True` | `newnode["internal"] = True` | ✅ |
| Literal CJK regex chars (style) | reverted to escaped `一-鿿` | ✅ |
| Deprecation `5.2` → `5.2.0` (trivial) | fixed | ✅ |
| Escape hatch untested under Sphinx | new `test_heading_anchors_html_ids_disabled_sphinx` | ✅ |
| Whitespace-slug skip only implicit | new `test_whitespace_slug_not_emitted_as_id` | ✅ |

**Left as-is, appropriately** — the points I'd flagged as optional: the cross-doc `_std_label_id_in_doc` linear scan (miss-path only), its name/id-precedence edge test, and a runtime `DeprecationWarning` on `default_slugify` (would be noisy). The #1073 `objects.inv` change was an awareness flag, not an action item.

Nice touch that the escape-hatch Sphinx test got the `1b920da` correction (asserting "no *extra* id" rather than slug absence, since for `# One` the slug equals the docutils id) — the test was clearly exercised rather than assumed. The one genuine latent fragility is now explicit and the two coverage gaps are closed; nothing outstanding from my side. 👍

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

---

PR #1175, opened 2026-07-20
🐛 Accurate directive line mapping and source-attributed nested renders

chrisjsewell:
Closes three directive-API fidelity gaps characterised while making sphinx-jinja2's warnings line-accurate (myst-parser 4.0.1/5.1.0, re-verified against master post-#1164).

## `content_offset` is now document-relative (rST convention)

Directives run under MyST received `content_offset = body_offset` (directive-relative), so the conventional `content_offset + N` idiom reported lines near the top of the file wherever the fence sat. Directives now receive the 0-based document line of their first content line, and `self.content.items` carry per-line absolute lines — both exactly what docutils' rST parser provides. Coupled with this, `MockState.nested_parse` treats `input_offset` as an absolute document offset. For the standard `nested_parse(self.content, self.content_offset, node)` idiom the rendered position is unchanged (the two old offsets summed to the same number), so **no rendering fixture changes**; the external contract is what's fixed.

Free corrections that fell out: `{epigraph}` attribution `.line` (was one too small), `{parsed-literal}` `.line` (was hardcoded 1), and myst's own `_docs.py` directives.

## Source-attributed nested renders (supported API)

`DocutilsRenderer.nested_render_text(text, lineno, source=...)` attributes rendered nodes **and** warnings (sphinx logger + docutils reporter, symmetrically) to a caller-specified path, exception-safe and re-entrant. Since #1114 switched warning locations to `env.docname`, the document/reporter override dance was inert under sphinx — `{include}` itself was affected: warnings inside included markdown were logged against the *including* document, one line off. Includes now cite the included file at the true line (the #1164 "tracked separately" off-by-one), and `:start-after:` no longer derives line numbers from a **character** index (previously past-EOF lines like `inc.md:19` in a 5-line file).

## Supported insertion API

- `MockStateMachine.insert_input(input_lines, source=None)` — docutils-compatible signature; parsed as MyST (documented caveat), rendered at the directive's position, attributed to `source:1..n`.
- `MockState.renderer` / `MockStateMachine.renderer` properties bless the access path.
- New `docs/develop/extending.md` documents the API with an explicit stability promise, `lineno` line-space semantics, eval-rst semantics (already document-absolute; now pinned by test), and limitations (deferred document-level warnings; inline tokens).

## Behaviour notes for extension authors (also in changelog)

- `nested_parse` with literal relative offsets (e.g. autodoc/autosummary's `offset=0`) now places content at the document top, as under rST.
- `self.lineno + self.content_offset` arithmetic now double-counts — as it always did under rST (sphinx-design's badges/buttons hit this latent pattern).
- Duplicate-reference-definition warnings still attribute to the containing document (deferred post-parse; documented limitation).

## Verification

- 1245 → **1285 passed**, 11 skipped; ruff/format/mypy clean; docs build clean.
- The originating scenario end-to-end: a fence at line 9 mapping an error on content line 2 via `content_offset + 2` now warns at `index.md:11` (was `index.md:3`).
- Master-vs-branch doctrees byte-identical on real docs; only warning-location outputs changed, each hand-verified against true source lines; parse-time delta within noise.
- Adversarially reviewed in two rounds (edge-case probes incl. nested/list/blockquote/eval-rst fences, include chains and slicing options, mutation testing of the new tests, downstream sweep: myst-nb unaffected, sphinx-design standard idioms invariant).

Downstream (sphinx-jinja2): the `isinstance(state_machine, StateMachine)` gate can become a version check, the override dance collapses onto `nested_render_text(..., source=template_path)`, and its `test_myst_render_error_location` can assert line-accurate expectations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,15 @@
 
 ### ✨ New Features
 
-- ✨ Add a supported `source=` argument to `DocutilsRenderer.nested_render_text`, so extensions that render generated content (an included file, a template's output) can attribute its nodes and warnings to the originating file or template, in both docutils and Sphinx builds
+- ✨ Add a `source=` argument to `DocutilsRenderer.nested_render_text`, so extensions rendering generated content (an included file, a template's output) can attribute its nodes and warnings to the originating file or template, in both docutils and Sphinx builds
 - ✨ Add `MockStateMachine.insert_input`, a docutils-compatible way for a directive to render generated MyST at its position
 - ✨ Expose the active renderer to directives via read-only `MockState.renderer` and `MockStateMachine.renderer` properties
 
 ### 🐛 Bug Fixes
 
-- 🐛 Make a directive's `content_offset` document-relative (the 0-based absolute line of its first content line), matching docutils' own reStructuredText parser, so extensions computing `content_offset + N` (e.g. sphinx-jinja2) report the correct line; `MockState.nested_parse` correspondingly treats its `input_offset` as an absolute document line. While here, `{epigraph}` attribution and `{parsed-literal}` node lines are corrected to their true document lines
-  - _Migration note for extension authors:_ `MockState.nested_parse`'s `input_offset` is now an absolute document offset (docutils-consistent). The standard `nested_parse(self.content, self.content_offset, node)` idiom is unchanged. However, a directive passing a literal *relative* offset (e.g. `0`) now gets docutils-identical placement — under Sphinx, autodoc/autosummary's default `offset=0` relocates rendered content to the top of the document, exactly as under reStructuredText — and arithmetic such as `self.lineno + self.content_offset` now double-counts the position (as it always did under reStructuredText)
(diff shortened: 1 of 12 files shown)


kdeldycke:
Thanks for this: the document-relative `content_offset` is exactly the contract I was missing.

One adjacent bug this PR doesn't cover, in case you'd like to fold it in while you're in here: `body_offset` is inflated by one when a directive has an option block **and** its body ends in blank line(s). It's independent of the relative-to-absolute reframe: it lives in the `len(content.splitlines()) - len(body_lines)` line, which this PR leaves unchanged, so it persists on this branch.

See:
```python
from docutils.parsers.rst.directives.admonitions import Note
from myst_parser.parsers.directives import parse_directive_text

parse_directive_text(Note, "", ":class: tip\nbody").body_offset        # 1  (correct)
parse_directive_text(Note, "", ":class: tip\nbody\n\n\n").body_offset  # 2  (should be 1)
# YAML-style option block inflates the same way: 3 -> 4
```

Root cause: `body_lines` comes back from `_parse_directive_options` through a `"\n".join(...)` round-trip that drops exactly one trailing blank line, so when the body ends in blank line(s) the subtraction over-counts the option block by one. Every body element is then reported one source line too low.

PR #1182, opened 2026-07-29
Apply safe modern typing and style fixes in myst_parser

inquilabee:
## Summary

Our analysis found **316 format issues** and about **286 refactoring opportunities** across the reviewed scope. Security scans found no leaked secrets; code duplication is low (~2%); maintainability index and cyclomatic complexity flagged a few hotspots (notably `mdit_to_docutils/base.py`); about 286 structural refactor suggestions remain. Lint volume is moderate for a package of this size.

The pull request is a **sample** of those findings — **10 files, ~55 focused changes** — not a full format pass or refactor cleanup.

Hotspots and improvement notes below are scoped to **Python (`*.py`)** source.

## What you are doing right

- **Security:** No secrets detected in scope (gitleaks clean).
- **Duplication:** Code duplication looks clean — jscpd reports ~1–2% across Python and other files.
- **Dependencies:** No known vulnerabilities in declared dependencies (pip-audit clean).
- **Tests & CI:** Strong pytest coverage (~91% on `myst_parser`) with a broad Python/Sphinx matrix in GitHub Actions.
- **Lint baseline:** Pre-commit with ruff + mypy on `myst_parser/` keeps the main package disciplined.

### Maintainability & complexity

- Radon flagged `myst_parser/mdit_to_docutils/base.py` at maintainability index rank C (MI=0.0) and P95 cyclomatic complexity 12 (ceiling 7) — the renderer module would benefit from staged extraction.

### Dependencies

- Deptry reported ~23 unused or missing optional dependency declarations — mostly dev/docs extras; classify before pruning.

## Changes

- `myst_parser/mdit_to_docutils/base.py` — `X | None` annotations, quoted `cast("str", …)` targets, set literal for slug token types.
- `myst_parser/mdit_to_docutils/sphinx_.py` — union syntax and minor style cleanups.
- `myst_parser/mdit_to_docutils/transforms.py` — union syntax.
- `myst_parser/inventory.py` — union syntax on optional types.
- `myst_parser/parsers/directives.py`, `options.py`, `parse_html.py`, `sphinx_.py` — union syntax and safe style fixes.
- `myst_parser/sphinx_ext/directives.py`, `mathjax.py`, `myst_refs.py` — union syntax and import/format tidy-ups.


myst_parser/sphinx_ext/mathjax.py
```diff
@@ -25,10 +25,10 @@ def log_override_warning(app: Sphinx, config_name: str, current: str, new: str)
     if logging.is_suppressed_warning("myst", "mathjax", app.config.suppress_warnings):
         return
     logger.warning(
-        f"`{config_name}['options']['processHtmlClass']` "
-        f"is being overridden by myst-parser: '{current}' -> '{new}'. "
-        "Set `suppress_warnings=['myst.mathjax']` to ignore this warning, or "
-        "`myst_update_mathjax=False` if this is undesirable."
+        "`%s['options']['processHtmlClass']` is being overridden by myst-parser: '%s' -> '%s'. Set `suppress_warnings=['myst.mathjax']` to ignore this warning, or `myst_update_mathjax=False` if this is undesirable.",
```
chrisjsewell: This collapses four readable lines into a single ~230-character string literal, well past the 88-column format width (ruff-format doesn't split string literals, so it passes silently). The lazy-`%`-formatting benefit is negligible for a warning emitted at most once per build, and the `G` (flake8-logging-format) rules aren't enabled in this repo anyway. The rendered message is unchanged (`test_mathjax_warning` still passes), but I'd suggest keeping the original multi-line f-string.

---

myst_parser/parsers/options.py
```diff
@@ -265,7 +265,9 @@ def _tokenize(text: str, state: State) -> Iterable[Token]:
         # find key
         ch = stream.peek()
         if ch in ("'", '"'):
-            yield _scan_flow_scalar(stream, cast(Literal['"', "'"], ch), is_key=True)
+            yield _scan_flow_scalar(
+                stream, cast("Literal['\"', \"'\"]", ch), is_key=True
```
chrisjsewell: This is where the `cast()`-quoting (TC006 style) turns into a net readability loss: `cast("Literal['\"', \"'\"]", ch)` with escaped quotes inside a string type expression is considerably harder to read than the original `cast(Literal['"', "'"], ch)`, and it forces line-wrapping churn here and in `mdit_to_docutils/sphinx_.py`. Since `TC` rules aren't enabled in this repo's ruff config, nothing enforces the quoted form going forward — it will just drift back. Suggest reverting the cast-quoting (here especially, and ideally throughout) unless TC006 is adopted in `pyproject.toml`.

---

myst_parser/inventory.py
```diff
@@ -11,6 +11,7 @@
 import argparse
 import functools
 import json
+import pathlib
```
chrisjsewell: The convention elsewhere in this package is `from pathlib import Path` (see `mdit_to_docutils/sphinx_.py`, `mocking.py`), so a bare `import pathlib` is inconsistent. More fundamentally, `pathlib.Path(uri).open("rb")` buys nothing over `open(uri, "rb")` when the value is only opened and never manipulated as a path — and the `PTH` rules aren't enabled in this repo's ruff config. Suggest reverting both hunks that make this substitution.

---

chrisjsewell (commented):
Thanks for the contribution. I verified this diff against the project's pinned toolchain — ruff 0.15.20 (`check` + `format --check`), mypy 2.1.0 with the repo config, and the full test suite (**1241 passed, 15 skipped**) — all green, and every control-flow rewrite is semantics-preserving (each converted `elif` follows a branch that unconditionally returns or raises). So the changes are technically safe. However, most of them enforce style conventions this repo has not adopted, and a few are readability regressions.

**The core issue: none of the changed patterns are backed by the project's ruff config.** `pyproject.toml` selects `B, C4, FA, FURB, I, ICN, ISC, N, PERF, PGH, PIE, RUF, SIM, UP` — not `TC` (quoted `cast()`), `RET` (`elif` after `return`), `PTH` (pathlib), or `G` (logging f-strings). Hand-fixing ~55 sampled sites means the style will silently regress and new code will continue the old pattern. If these conventions are wanted, the right vehicle is enabling the corresponding rules in `pyproject.toml` plus a repo-wide `ruff check --fix` in one atomic PR; if not, this diff is churn with git-blame cost.

Notes per change category:

- **Union reordering** (`None | X` → `X | None`): harmless, though the codebase already uses PEP 604 unions throughout — this only flips operand order, which no ruff rule enforces, so "modernized union syntax" oversells it.
- **`elif` → `if` after return/raise, `count += 1`, set-literal membership**: fine, verified equivalent.
- **Quoted `cast()` targets**: `TC006` isn't enabled, and in `parsers/options.py` it produces escaped-quote literals harder to read than what they replace (inline comment).
- **`pathlib.Path(uri).open()`**: `PTH` isn't enabled, `import pathlib` deviates from the repo's `from pathlib import Path` convention, and there's no benefit when the value is only opened (inline comment).
- **`mathjax.py` logging rewrite**: message renders identically (covered by `test_mathjax_warning`), but it collapses four readable lines into one ~230-char string literal (inline comment).

Two mislabels in the PR description worth noting: quoting `cast()` targets is `TC006` (`UP037` *removes* quotes), and the list→set membership change isn't `UP038` (the deprecated isinstance-tuple rule).

**Recommendation:** trim the PR to the neutral subset (union reordering, `elif` → `if`, `count += 1`, set literal) and revert the mathjax, quoted-cast, and pathlib hunks — or close in favour of a config-level decision on which ruff rules to enable.

---

inquilabee:
Thanks @chrisjsewell — that's a fair and thorough review.

I've pushed a follow-up that trims the PR to the neutral subset you suggested:

- **Kept:** union operand reordering (`X | None`), `elif` → `if` after early return/raise, `count += 1`, and set-literal membership.
- **Reverted:** `pathlib.Path(...).open()` in `inventory.py`, quoted `cast()` targets, and the `mathjax.py` logging rewrite.

Agreed that enabling additional ruff rules repo-wide is the right vehicle for the rest — not hand-edited sample diffs.

For context on the tooling side: [ShipGate](https://github.com/inquilabee/shipgate) separates **auto refactors** (`refactor fix`, safe rules applied directly) from **hint-tier suggestions** (`refactor check --strict`, for a human or agent to apply manually). Examples from this repo that fall into the hint/auto bucket include `RET506` (drop redundant `elif` after return), `SIM118`/`C419`-style set membership, union reordering, and optional rules like `PTH123` / `TC006` when they're not in your config.

Happy to close if you'd rather land only a config-level decision — appreciate the time either way.

inquilabee:
Thanks for the thorough review, @chrisjsewell. You're right that most of these changes enforce style conventions the project hasn't adopted — I don't think another push is the right fit. Closing this PR; appreciate you verifying the toolchain and the inline notes.

